PR #6, opened 2025-06-04
Update README.md

harrycis:
The env doesn't guarantee to be passed along during subprocess sprawl. This can be verified by Claude Desktop

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,14 +100,18 @@ Open your `claude_desktop_config.json` and add the mcp server to the list of `mc
         "-i",
         "--rm",
         "-e",
-        "SLACK_MCP_XOXC_TOKEN=xoxc-...",
+        "SLACK_MCP_XOXC_TOKEN",
         "-e",
-        "SLACK_MCP_XOXD_TOKEN=xoxd-...",
+        "SLACK_MCP_XOXD_TOKEN",
         "ghcr.io/korotovsky/slack-mcp-server",
         "mcp-server",
         "--transport",
         "stdio"
-      ]
+      ],
+      "env": {
+        "SLACK_MCP_XOXC_TOKEN": "xoxc-...",
+        "SLACK_MCP_XOXD_TOKEN": "xoxd-..."
+      }
     }
   }
 }
--- a/README.md
+++ b/README.md
@@ -100,18 +100,14 @@ Open your `claude_desktop_config.json` and add the mcp server to the list of `mc
         "-i",
         "--rm",
         "-e",
-        "SLACK_MCP_XOXC_TOKEN=$SLACK_MCP_XOXC_TOKEN",
+        "SLACK_MCP_XOXC_TOKEN=xoxc-...",
         "-e",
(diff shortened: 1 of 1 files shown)


korotovsky:
Hi @harrycis, ~it seems it's a bug in Claude Desktop (see here: https://github.com/anthropics/claude-code/issues/1254) so I suggest to fix README a bit differently, keep env (because it was definitely working for some versions)add a warning section~

harrycis:
Actually, I was wrong. Claude does pass it to its child process but docker doesn't.

So it is not a bug of Claude ~but just how docker works~. The env has been passed to docker, but docker doesn't pass the env to the container it launches.

You can verify it by this setup:
1. create a script with a single line `env >/tmp/env.txt`. Let's say it is saved as `env.sh`
2. add an mcp entry pointing to the script
```
"test-env": {
      "command": "bash",
      "args": [
       "/tmp/env.sh"
      ],
      "env": {
        "SLACK_MCP_XOXC_TOKEN": "abcd"
      }
    }
```
3. Restart Claude Desktop and you can see `SLACK_MCP_XOXC_TOKEN` has been passed to it.

korotovsky:
Hi @harrycis it's me again, I decided to dig into the issue a little and thank you for bringing it up. So here is the summary:

tl;dr: everything works with the following config:
<details>

```json
{
  "mcpServers": {
    "slack": {
      "command": "docker",
      "args": [
        "run",
        "-i",
        "--rm",
        "-e",
        "SLACK_MCP_XOXC_TOKEN",
        "-e",
        "SLACK_MCP_XOXD_TOKEN",
        "ghcr.io/korotovsky/slack-mcp-server",
        "mcp-server",
        "--transport",
        "stdio"
      ],
      "env": {
        "SLACK_MCP_XOXC_TOKEN": "xoxc--...",
        "SLACK_MCP_XOXD_TOKEN": "xoxd-..."
      }
    }
  }
}
```
</details>

The difference is we don't pass explicitly values to the `-e` opt anymore.

Here is the screenshot:

photo_5337277686495704929_y

Could you please update PR according to the finding and also test it on your end?

korotovsky:
Not sure it this is also working option or a typo 

but in my last example it was

```
...
"SLACK_MCP_XOXC_TOKEN",
...
```
instead of 

```
...
"SLACK_MCP_XOXC_TOKEN=SLACK_MCP_XOXC_TOKEN",
...
```

harrycis:
Yea, that works. It was likely that it was passed as shell variable to construct the docker args, but it isn't a shell sprawl by Claude. PR updated

harrycis:
> Not sure it this is also working option or a typo
> 
> but in my last example it was
> 
> ```
> ...
> "SLACK_MCP_XOXC_TOKEN",
> ...
> ```
> 
> instead of
> 
> ```
> ...
> "SLACK_MCP_XOXC_TOKEN=SLACK_MCP_XOXC_TOKEN",
> ...
> ```

My bad, going to fast.

korotovsky:
Thank you! proper README is very important for project adoption.

PR #10, opened 2025-06-06
more flexible params for channels_list

harrycis:
In real real-world setup, the channel list can be large and exceed the model context size limitation. Need to be able to break it down into batches.

### DIFF
--- a/pkg/handler/channels.go
+++ b/pkg/handler/channels.go
@@ -22,7 +22,7 @@ type Channel struct {
 	Topic       string `json:"topic"`
 	Purpose     string `json:"purpose"`
 	MemberCount int    `json:"memberCount"`
-	Cursor   	string `json:"cursor"`
+	Cursor      string `json:"cursor"`
 }
 
 type ChannelsHandler struct {
@@ -76,10 +76,10 @@ func (ch *ChannelsHandler) ChannelsHandler(ctx context.Context, request mcp.Call
 		Types:           channelTypes,
 		Limit:           limit,
 		ExcludeArchived: true,
-		Cursor:			 cursor,
+		Cursor:          cursor,
 	}
 	var (
-		total int
+		total   int
 		nextcur string
 	)
 	for {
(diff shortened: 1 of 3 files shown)


pkg/handler/channels.go
```diff
@@ -54,46 +55,76 @@ func (ch *ChannelsHandler) ChannelsHandler(ctx context.Context, request mcp.Call
 		}
 	}
 
+	cursor := ""
+	cursorOpt := request.Params.Arguments["cursor"]
```
korotovsky, line 59: Hi, I have recently updated MCP framework and now the new API is available 

```go
	sortType := request.GetString("sort", "popularity")
	types := request.GetString("channel_types", PubChanType)
```

Could you please apply this new api for `cursor` and `limit` ? Also please make sure you have the latest master.

harrycis: Thanks. Refactored.

pkg/handler/channels.go
```diff
@@ -54,46 +56,65 @@ func (ch *ChannelsHandler) ChannelsHandler(ctx context.Context, request mcp.Call
...
-	for i := 1; ; i++ {
-		var (
-			chans   []slack.Channel
-			nextcur string
-		)
+	var (
+		total int
+		nextcur string
+	)
+	for {
```
korotovsky, line 85: Could you please explain what is the reason to use loop here if we supply `cursor` for subsequent separate HTTP queries? From what I understood, the logic is similar to cursor pagination from the conversations, but there it is only one request per `limit` or per `cursor`, Claude Desktop initiates itself the next one and supplies the cursor if needed.

harrycis: Per the Slack API [doc](https://api.slack.com/apis/pagination#facts), a single call may return fewer records than the specified limit.  If the semantic at the MCP layer is to fetch up to `limit` number of records, multiple calls might be needed.

korotovsky (approved):
:rocket:

Issue #16, opened 2025-06-13
[FeatureRequest] Set username for direct messages ('im' type)


orinciog:
The answer from `channels_list` for direct messages is not very helpful. It contains only the ID of the conversation, but without any information about with whom is that conversation. 

<img width="610" alt="Image" src="https://github.com/user-attachments/assets/e287d598-9dc8-4042-8365-6b4c9119cdf7" />

Using [conversation.list](https://api.slack.com/methods/conversations.list) method from slack api, the response for the one 'im' channel is the following (it has "user" information).

 ```
{
            "id": "D08",
            "created": 1748938480,
            "is_frozen": false,
            "is_archived": false,
            "is_im": true,
            "is_org_shared": false,
            "context_team_id": "",
            "updated": 1748938480459,
            "user": "U05R",
            "is_user_deleted": false,
            "priority": 0.22945248020388
        }
```

It will be very helpful if you can augment the information for this type of conversation with the realName of the user? (just like you did for the messages). (Maybe you can put this information in the `Name` field: `private conversation with user realName` instead of that `#` ).

It will be very helpful because after that, using this mcp one can respond very quickly at questions like : "Summarize my discussions with user USER". 

Also, I can submit a PR if you want for this feature.

Thank you

korotovsky:
Hi @orinciog, thanks for the feedback!

Submitting is always very welcome as it helps me a lot as at the moment I'm trying to dig into the authentication hiccups in some cases of some users.

PR #19, opened 2025-06-16
For direct messages, display username in conversation list

orinciog:
This PR resolves https://github.com/korotovsky/slack-mcp-server/issues/16


korotovsky (commented):
Thank you! As a part of this PR may I ask you to check if DM with multiple participants will be correctly displayed as well? I think it also needs a fix. Would be nice to have complete fix at once on this topic.

orinciog:
A DM with multiple participants is actually a mpim object.
This object has `name` and `purpose.value` keys. These keys contain the name of participants in a mpim. 

You can test it [here](https://api.slack.com/methods/conversations.info/test). 

I modified my PR not to consider mpim conversations. Thank you for your attention.

PR #23, opened 2025-06-24
feat: add SLACK_MCP_XOXP_TOKEN user OAuth authentication support

HeroSizy:
## Summary
- Add support for User OAuth tokens (xoxp-) as an alternative authentication method
- Maintain full backward compatibility with existing XOXC/XOXD session-based authentication  
- XOXP tokens take priority when both authentication methods are available
- Update demo mode detection to support both authentication methods

## Changes Made
- **pkg/provider/api.go**: Added `newWithXOXP()` and `newWithXOXC()` functions with fallback logic
- **cmd/slack-mcp-server/main.go**: Updated demo mode detection for both token types
- **README.md**: Added comprehensive User OAuth setup documentation and configuration examples
- **CLAUDE.md**: Updated environment variables documentation

## Benefits
- More secure authentication option that doesn't require browser session extraction
- Follows OAuth best practices with proper user tokens
- Zero breaking changes for existing users
- Clear migration path from session-based to OAuth authentication

## Test Plan
- [x] Code compiles successfully
- [x] Maintains backward compatibility with XOXC/XOXD tokens
- [x] XOXP token authentication works with standard OAuth flow
- [x] Demo mode works with both authentication methods
- [x] Documentation is comprehensive and accurate

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,26 @@ Token value is printed right after the executed command (it starts with
  - Press Ctrl+C or Cmd+C to copy it's value to clipboard.
  - Save it for later.
 
+#### Alternative: Using `SLACK_MCP_XOXP_TOKEN` (User OAuth)
+
+Instead of using browser-based tokens (XOXC/XOXD), you can use a User OAuth token:
+
+1. Go to [api.slack.com/apps](https://api.slack.com/apps) and create a new app
+2. Under "OAuth & Permissions", add the following scopes:
+   - `channels:history` - View messages in public channels
+   - `channels:read` - View basic information about public channels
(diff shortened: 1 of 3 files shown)


korotovsky (commented):
Thanks for the great contribution! 

You might notice, this Slack MCP has been on purpose designed to authenticate via `xoxc` and `xoxd` pairs to be able to work "in stealth mode", so you would not need to neither create bots nor approve it.

So in context and ideology of this repository I'd personally not call it "recommended" way, although I agree that OAuth is more secure.

HeroSizy:
Got you, will remove the recommended part

PR #36, opened 2025-07-04
New tools: Add Message & Search

korotovsky:
This PR brings two more tools: `conversations_add_message` and `conversations_search_messages`


Anima-t3d:
for `conversations_add_message` there is an allowed list, would it be useful to have an `excluding_list`? To explicitly at all times reject writing message in such channel/group? Meaning if the channel is in both allow and exclude, exclude should take precedence.

I think it's a good idea to explicitly allow the write.. but use case might be for user who want to allow all except certain sensitive channels? If user has 100 channels, need to allow 99 channels, although only need 1 channel to be excluded?

korotovsky:
Hi @Anima-t3d, thanks for the review. I have added negate logic, so the env variable remains the same but now it is possible either to define `C1234567890` or to exclude `!C1234567890`. This syntax must be quite straightforward in my opinion.

Defining both (mix) does not make sense in my opinion because then you would rather need to stick to one of the sides either all negate via `!` or via explicit whitelist.

danrasmuson:
@korotovsky -- I'm really loving slack-mcp. Its a great addition. My take is that slack-mcp ought to have full permission by default because this is easiest for users to get up and running and do what they want. Right now, from what I understand, we need to set `SLACK_MCP_ADD_MESSAGE_TOOL = 'true'` before slack-mcp has permission to post messages on our behalf. I'm challenging this decision because it feels like undue friction in a plugin that is badass to start with.

korotovsky:
Hi @danrasmuson,

The `SLACK_MCP_ADD_MESSAGE_TOOL` environment variable is a simple safeguard to protect existing pipelines and workflows that were already using the Slack MCP Server before this new message-posting tool was introduced.

When the tool was added in a recent update, I wanted to make sure that older workflows wouldn’t suddenly start posting messages in unexpected or unsafe ways. This variable helps prevent that by requiring an explicit opt-in.

Since you already need to set up `XOXP/C/D_*` tokens anyway, it should be easy to also configure this variable to enable the message-posting feature safely.

Thanks for warm feedback!

PR #39, opened 2025-07-09
(fix) Fixed environment variables for host and port and default values

failattu:
The current documentation had a mistake of old variable naming, which no longer exists in source code and has been patched in docker compose files, but not in documentation, so fixed the documentation.

This caused me some headache and required reading the source code so would help people whom read the documentation.


korotovsky:
Hi @failattu, I'm very sorry for that! Thanks for the contribution.

failattu:
Open source project, not a problem :) that's why I fixed it for future

PR #65, opened 2025-07-24
Fix: Redirect console logs to stderr for stdio transport compatibility

clement-izard:
The server was outputting structured JSON logs to stdout when using stdio transport, breaking JSON-RPC 2.0 protocol compliance with MCP clients like Claude Code.

Root cause: zap logger configured with OutputPaths: ["stdout"] was sending all diagnostic logs to stdout, interfering with JSON-RPC messages that MCP clients expect exclusively on stdout.

Solution: Redirect all console logs to stderr while preserving stdout exclusively for JSON-RPC protocol communication.

Fixes compatibility with all MCP clients using stdio transport including Claude Code CLI, MCP Inspector, and other implementations.

### DIFF
--- a/cmd/slack-mcp-server/main.go
+++ b/cmd/slack-mcp-server/main.go
@@ -25,7 +25,7 @@ func main() {
 	flag.StringVar(&transport, "transport", "stdio", "Transport type (stdio or sse)")
 	flag.Parse()
 
-	logger, err := newLogger()
+	logger, err := newLogger(transport)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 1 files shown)


korotovsky:
Hi @clement-izard, thank you for the fix! Indeed I overlooked it. But should we only do this only for `stdio` transport not not affect `sse` ?

clement-izard:
Hey! That's right!

I updated the code, it will use the correct output now. Thanks for having a look!

Edit: Regarding failing tests it's because we're missing the env variables, it shoudn't be related with this code.

korotovsky:
@clement-izard thank you! May I ask you to merge with master and remove self-explaining comment in code. After I will merge asap.

Regarding tests all envs are in place, it was the issue in workflow configuration how to access secrets from forks. It should be fine now with latest fixes from `master`.

clement-izard:
For sure! I'll do that after lunch, count ~3h

clement-izard:
@korotovsky All good!

korotovsky:
Thank you!

PR #72, opened 2025-08-01
Fix Enterprise Grid xoxp token support and rate limiting errors

dionny:
## Summary
This PR fixes two critical issues:
1. **Enterprise Grid xoxp token support** - Fixed incorrect API selection for xoxp tokens in Enterprise Grid workspaces
2. **Rate limiting errors** - Fixed rate limiter implementation to properly prevent Slack API rate limit errors

### Problem
When using xoxp tokens (user tokens) in Enterprise Grid workspaces, the application was incorrectly attempting to use Enterprise Grid Edge APIs, which don't support xoxp tokens. This caused failures when fetching channels and users.

### Root Cause
The code was only checking if the workspace was Enterprise Grid (`isEnterprise`) but not considering the token type. Enterprise Grid Edge APIs only work with xoxc tokens, not xoxp tokens.

### Solution
- Added token type detection to check if the token starts with "xoxp-"
- Added `useEdgeAPI` flag that is only true when BOTH conditions are met:
  - Workspace is Enterprise Grid (`isEnterprise = true`)
  - Token is NOT an xoxp token (`\!isXOXPToken`)
- Updated `GetConversationsContext` to check `useEdgeAPI` instead of just `isEnterprise`
- Added proper fallback in `ClientUserBoot` for xoxp tokens

### Problem
The application was experiencing "slack rate limit exceeded, retry after 30s" errors when fetching channels and users.

### Root Cause
1. Rate limiter was being called AFTER making API requests instead of BEFORE
2. Each function created its own rate limiter instance, so concurrent API calls weren't coordinating

### Solution
1. Moved all `lim.Wait(ctx)` calls to happen BEFORE API calls
2. Added a shared `rateLimiter` field to `ApiProvider` struct
3. All API methods now use the same rate limiter instance
4. Using conservative `Tier2` limiter (3 seconds between requests, burst of 3) to stay well below Slack's limits

## Changes Made
- `pkg/provider/api.go`:
  - Added `useEdgeAPI` field and token type detection logic
  - Fixed API selection logic in `GetConversationsContext` and `ClientUserBoot`
  - Added shared `rateLimiter` field to `ApiProvider`
  - Fixed rate limiter calls to happen before API requests
- `pkg/limiter/limits.go`:
  - Added `Tier2` configuration for conservative rate limiting

## Test plan
- [x] Build completes successfully
- [x] Tested with xoxp token in Enterprise Grid workspace - channels and users fetch successfully
- [x] Verified no more rate limit errors with large workspaces
- [ ] Test with xoxc token to ensure Edge API still works correctly
- [ ] Test with non-Enterprise Grid workspace

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .idea
 .env
 .users_cache.json
+.channels_cache.json
 /extension.dxt/server/slack-mcp-server-*
 /extension.dxt/server/index.js
 /build/slack-mcp-server-*
--- a/pkg/limiter/limits.go
+++ b/pkg/limiter/limits.go
@@ -19,7 +19,7 @@ func (t tier) Limiter() *rate.Limiter {
 
 var (
 	// tier1 = tier{t: 1 * time.Minute, b: 2}
-	// tier2 = tier{t: 3 * time.Second, b: 3}
+	Tier2      = tier{t: 3 * time.Second, b: 3}
 	Tier2boost = tier{t: 300 * time.Millisecond, b: 5}
 	Tier3      = tier{t: 1200 * time.Millisecond, b: 4}
 	// tier4      = tier{t: 60 * time.Millisecond, b: 5}
--- a/pkg/provider/api.go
+++ b/pkg/provider/api.go
@@ -128,10 +128,10 @@ func NewMCPSlackClient(authProvider auth.Provider, logger *zap.Logger) (*MCPSlac
 	token := authProvider.SlackToken()
 	isXOXPToken := strings.HasPrefix(token, "xoxp-")
 	isEnterprise := authResp.EnterpriseID != ""
-	
+
 	// Only use Edge API if we're in Enterprise Grid AND not using an xoxp token
 	useEdgeAPI := isEnterprise && !isXOXPToken
-	
+
 	// Log token type and API selection for debugging
 	logger.Debug("Token type detection",
 		zap.Bool("isXOXPToken", isXOXPToken),
--- a/pkg/provider/api.go
+++ b/pkg/provider/api.go
@@ -77,6 +77,7 @@ type MCPSlackClient struct {
 	authProvider auth.Provider
 
 	isEnterprise bool
+	useEdgeAPI   bool // Controls whether to use Enterprise Grid APIs
 	teamEndpoint string
 }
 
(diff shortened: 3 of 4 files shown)


dionny:
The unit tests are likely failing because the workflow uses pull_request_target instead of pull_request.  I'm able to run the unit tests locally.

korotovsky:
Hi @dionny, thanks for trying to approach API provider, the core of this MCP Server - the most complicated part! I'd like to clarify a couple of questions before we get to code review.

Generally this MCP has the following logic when I or you should test it:

Every method must work with both types of tokens `xoxc`/`xoxd` and `xoxp` for every "feature" on both normal Slack and Enterprise Slack - 2 x 1 x 2. So that means channels refresh and users refresh we would need to "test" 8 use cases.

| Token Type     | Slack Environment   | Feature           | Test Case ID |
|----------------|---------------------|--------------------|--------------|
| `xoxc`/`xoxd`   | Normal Slack        | Channels Refresh   | TC-01        |
| `xoxc`/`xoxd`   | Normal Slack        | Users Refresh      | TC-02        |
| `xoxc`/`xoxd`   | Enterprise Slack    | Channels Refresh   | TC-03        |
| `xoxc`/`xoxd`   | Enterprise Slack    | Users Refresh      | TC-04        |
| `xoxp`          | Normal Slack        | Channels Refresh   | TC-05        |
| `xoxp`          | Normal Slack        | Users Refresh      | TC-06        |
| `xoxp`          | Enterprise Slack    | Channels Refresh   | TC-07        |
| `xoxp`          | Enterprise Slack    | Users Refresh      | TC-08        |

`ClientBoot` is part of `RefreshUsers` it narrows down us to 4 test cases, two with `xoxc`/`xoxd` pair and two with `xoxp`. Could you please provide what does client boot returns for you when you use `xoxp` on your enterprise configuration using HttpToolkit Mitm (please see env vars it is easy to enable) and ou can download it here: https://httptoolkit.com/

/cc @orinciog putting you in cc as you were interested of approaching it too.

korotovsky:
Closed in favor of smaller PR https://github.com/korotovsky/slack-mcp-server/pull/84/files after weeks of investigations. Some changes were cherry-picked. Thank you for the efforts.

PR #91, opened 2025-08-21
feat: add reactions payload to messages

koladev32:


korotovsky:
Hi @koladev32, thanks for the PR. Could you show how does it look as per raw CSV response with emojis?

koladev32:
> Hi @koladev32, thanks for the PR. Could you show how does it look as per raw CSV response with emojis?

@korotovsky here's an example of what it returns in CSV.

```csv
MsgID,UserID,UserName,RealName,Channel,ThreadTs,Text,Time,Reactions,Cursor
17557xxxx.3223xxx9,U02MTDASNCC,kolawole.mangabo,kolawole.mangabo,C09XXXXXLV,,Leads,2025-08-21T10:11:57Z,"[{""name"":""white_check_mark"",""count"":1,""users"":[""U02XXXXNCC""]}]",
```

korotovsky:
Thank you for output. As you might see, the flat-CSV format has been selected exactly for the reason to save LLM-tokens for end-users. There is a good example in your output how user data is being handled, I think if we agree on adding emoji then it must follow the same idea.

koladev32:
> Thank you for output. As you might see, the flat-CSV format has been selected exactly for the reason to save LLM-tokens for end-users. There is a good example in your output how user data is being handled, I think if we agree on adding emoji then it must follow the same idea.

@korotovsky I can see that. The difficulty here I will say is how the Reactions object from Slack is structured, but if we focus on showing the LLM the reactions only, then we can just retrieve the reaction name, and the number of occurrence of that reaction.

I don't have a straight idea yet, but I can see something like this `emoji:count|emoji:count` (`white_check_mark:3|thumbsup:8|heart:2|rocket:1`) but I'm afraid that without context or description, an LLM will have issues interpreting this format.

korotovsky:
Not sure that we would need additional tool prompting, I just tested it with ChatGPT your suggestion, so it might work:

Screenshot from 2025-08-21 21-05-55

koladev32:
> Not sure that we would need additional tool prompting, I just tested it with ChatGPT your suggestion, so it might work:
> Screenshot from 2025-08-21 21-05-55

@korotovsky  I made some tests with Gpt4, Gpt4, Claude Sonnet 4.1 and Gemini 2.5 and they understand the format pretty well. I made the changes.

Here is a an example message: 

```csv
MsgID,UserID,UserName,RealName,Channel,ThreadTs,Text,Time,Reactions,Cursor
1755xxx1.53xxx79,U0xx2MxxxxxNCC,kolawole.mangabo,kolawole.mangabo,C09xxxxxV,,This is dope for a lead.,2025-08-21T21:38:01Z,eyes:1|white_check_mark:1|raised_hands:1|smile:1,
```

korotovsky:
Thank you! One final thing - please could you remove self-elaborating comments from the PR (vibe-auto-generated?).

koladev32:
> Thank you! One final thing - please could you remove self-elaborating comments from the PR (vibe-auto-generated?).

@korotovsky I removed them

korotovsky:
Thanks! 

I checked it the Slack search with emojis and Slack actually returns them there:

Screenshot from 2025-08-22 09-35-14

Have you tried to investigate on this?

koladev32:
> Thanks!
> 
> I checked it the Slack search with emojis and Slack actually returns them there:
> Screenshot from 2025-08-22 09-35-14
> 
> Have you tried to investigate on this?

@korotovsky Yes. The fact we can't see the reactions on the search messages is due to the way the search data is converted to match the `Message` type in the project. 

1. The search query retrieve results that comes structured in the type ` slack.SearchMessage`. This type does not have reactions fields attached to it. 
```go
type SearchMessage struct {
	Type        string       `json:"type"`
	Channel     CtxChannel   `json:"channel"`
	User        string       `json:"user"`
	Username    string       `json:"username"`
	Timestamp   string       `json:"ts"`
	Blocks      Blocks       `json:"blocks,omitempty"`
	Text        string       `json:"text"`
	Permalink   string       `json:"permalink"`
	Attachments []Attachment `json:"attachments"`
	Previous    CtxMessage   `json:"previous"`
	Previous2   CtxMessage   `json:"previous_2"`
	Next        CtxMessage   `json:"next"`
	Next2       CtxMessage   `json:"next_2"`
}
```
This appears to be an implementation gap in the `slack-go/slack` Go package, as browser inspection shows the Slack API actually returns reactions data in `search.messages responses`, but the Go library's `SearchMessage` struct doesn't include the Reactions field to capture it.
2. Because that field does not exist, we set `Reactions` to an empty string.

PR #102, opened 2025-09-24
Implement Streamable HTTP transport

### LINKED ISSUE #87: Add support for Streamable HTTP
Streamable HTTP is a replacement for HTTP+SSE ([ref](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http)) and the latest and greatest MCP protocol. It is better suited for multi-user, cloud based clients and tools like [LibreChat](https://www.librechat.ai/docs/features/mcp#server-transports) recommend it as the only suitable option for production use. This seems to be the most feature-rich and well-maintaned Slack MCP server on GitHub and it would be great to see it support this newer transport.

Graf2242:
I have implemented http transport as another option, keeping stdio and sse untouched.
Rename SLACK_MCP_SSE_API_KEY to SLACK_MCP_API_KEY and now it's used in sse and http both
Bump mcp-go version, there was some critical for streaming http changes

Update documentation

Closes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Slack MCP Server
 [![Trust Score](https://archestra.ai/mcp-catalog/api/badge/quality/korotovsky/slack-mcp-server)](https://archestra.ai/mcp-catalog/korotovsky__slack-mcp-server)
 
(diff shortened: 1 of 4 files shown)


korotovsky (commented):
Hi @Graf2242, that's truly awesome contribution! ~One thing: may I ask you to keep `SLACK_MCP_SSE_API_KEY`  for users who continue use SSE? Otherwise this feature would be a breaking change for them.~

PR #107, opened 2025-10-03
Fix filter by private channel id, and private channel name

wllmtrng:
Note: Please check against newer Slack API Specs. This is how my company's slack api returns. 

### Problem
When using the `conversations_search_messages` tool with the `filter_in_channel` parameter, providing a channel name with a `#` prefix (e.g., `#general`) resulted in a malformed Slack search query with a double hash: `in:##general`, causing zero search results.

**Example of broken request:**
```json
{
  "filter_in_channel": "#general",
  "search_query": "meeting"
}
```

**Resulted in invalid Slack query:**
```
"meeting in:##general"
```

### Root Cause
The `paramFormatChannel` function was returning the channel name with a `#` prefix, which was then concatenated with the `in:` filter by `buildQuery`, resulting in `in:#channel-name`. However, when the user already provided the `#` in their input, this created a double `##`.

### Solution
- Modified `paramFormatChannel` to return the channel name **without** the `#` prefix
- The `buildQuery` function then correctly formats it as `in:channel-name` 
- Added support for `G` prefix (private groups/channels) in addition to existing `C` prefix (standard channels)

**Now correctly generates:**
```
"meeting in:general"
```

### Changes
- Updated `paramFormatChannel()` to strip `#` from return value
- Added support for both `C` and `G` channel ID prefixes
- Simplified comment to clarify the function's behavior

### Testing
- ✅ Builds successfully
- ✅ Handles `#channel-name` format
- ✅ Handles `C123456789` channel ID format  
- ✅ Handles `G123456789` private group ID format

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -713,13 +713,14 @@ func (ch *ConversationsHandler) paramFormatChannel(raw string) (string, error) {
 	cms := ch.apiProvider.ProvideChannelsMaps()
 	if strings.HasPrefix(raw, "#") {
 		if id, ok := cms.ChannelsInv[raw]; ok {
-			return "#" + cms.Channels[id].Name, nil
+			return cms.Channels[id].Name, nil
 		}
 		return "", fmt.Errorf("channel %q not found", raw)
 	}
-	if strings.HasPrefix(raw, "C") {
+	// Handle both C (standard channels) and G (private groups/channels) prefixes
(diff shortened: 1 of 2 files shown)


pbassut:
Was this released?

korotovsky:
Yes, here https://github.com/korotovsky/slack-mcp-server/releases/tag/v1.1.26

PR #116, opened 2025-10-17
Add slack manifest to XOXP configuration

mattppal:
This PR adds a valid manifest configuration to the XOXP instructions.

This should reduce user error and speed up the install / config process for XOXP


korotovsky:
Hi @mattppal, thanks for the docs improvements!

PR #121, opened 2025-10-29
Updating the ConversationsMessageAddHandler to have optional attachments

chongzluong:
# Overview

For the purposes of [a demo video](https://x.com/chongzluong/status/1983220353813107052) that utilizes Slack MCP and [Cartesia MCP](https://github.com/cartesia-ai/cartesia-mcp/tree/main), I needed to be able to attach files to messages through MCP, so I made the following changes:

## Slack API

- Added [File Upload](https://docs.slack.dev/reference/methods/files.upload/#markdown)
- Added [Open Conversation](https://docs.slack.dev/reference/methods/conversations.open/)

## MCP

- Updated the `addMessageParams` to include a potential `attachmentURL` and `attachmentName`
  - (Optional) `attachmentURL` can be a local file or publicly accessible file URL. The file at this URL will be downloaded in and used as the body for the file upload API
  - (Optional) `attachmentName` the name to use for the uploaded file
- Updated the `ConversationsAddMessageHandler` to incorporate a file attachment path
  - If an `attachment_url` is specified, we interpret local vs remote files , download, and then call the Slack API for uploading files (with an optional initial comment + thread timestamp)
  - If no `attachment_url` is specified, the previous behavior of calling the Slack API to post a message works as it did before

# Testing

The tweet linked above exhibits the functionality in action.
1. I ran `make build` to build the binary
2. I pointed the Claude Desktop config to the binary with the `--transport stdio` arguments and the relevant Slack tokens
3. I ran Cartesia's MCP
4. I had it generate me a script, TTS'd the script into an audio file using Cartesia's MCP, and then sent it to myself afterwards directly over slack. Screenshot below:

Screenshot 2025-10-28 at 10 24 38 PM

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -5,8 +5,10 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -73,10 +75,12 @@ type searchParams struct {
 }
 
 type addMessageParams struct {
-	channel     string
-	threadTs    string
-	text        string
-	contentType string
+	channel        string
+	threadTs       string
+	text           string
+	contentType    string
+	attachmentUrl  string
+	attachmentName string
 }
 
 type ConversationsHandler struct {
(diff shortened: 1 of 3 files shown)


chongzluong:
@korotovsky apologies I'm under the impression that the integration test failures above aren't related to my changes, but please let me know if I'm incorrect there 😅

korotovsky:
Hi @chongzluong, thank you for submitting great functionality for this Slack MCP. However the security aspect of the PR in my opinion must be improved a little before we merge it to upstream, because this is the most popular Slack MCP Server nowadays and we should take it very serious especially when we would like to add such things like file upload/etc. let me summarize some points what is missing and should be added:

- Env var that makes file uploading disabled by default, similar to posting messages env so all existing deployments won't be affected by this feature.
- Scoping of where MCP can access files. It might be a comma separated list of patters on local filesystem i.e. /Users/*,/usr/data or comma separated list of domains where MCP is allowed to download files from (similar to link unfurling protection)

chongzluong:
@korotovsky definitely makes sense, I'll update the PR later this week and follow up

korotovsky:
@chongzluong ping

chongzluong:
@korotovsky ah apologies, it's been quite a busy last month and a half on my end. I will try to find some time to tackle this on Sunday if that's alright

PR #127, opened 2025-11-12
docs: clarify DM channel IDs and user mention formatting

nateGeorge:
This PR includes two improvements to DM and group DM support:

1. **Documentation**: Clarifies that DM messages require channel IDs starting with D (not @username format) and documents how to mention users with <@USER_ID> format. Includes examples for finding DM channel IDs and user IDs.

2. **Feature**: Adds userId, userName, members, and memberNames fields to the channels_list CSV output for IM and MPIM channels, making it easy to identify which user(s) each DM/group DM is with. This resolves the issue where DM channels were listed without any indication of who they were with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,42 @@ Add a message to a public channel, private channel, or direct message (DM, or IM
 > **Note:** Posting messages is disabled by default for safety. To enable, set the `SLACK_MCP_ADD_MESSAGE_TOOL` environment variable. If set to a comma-separated list of channel IDs, posting is enabled only for those specific channels. See the Environment Variables section below for details.
 
 - **Parameters:**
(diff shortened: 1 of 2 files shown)


korotovsky (changes requested):
Hi @nateGeorge, thank you for the PR, I think it would make sense instead of adding 4 more properties into `Channel` struct to add real name of the IM or MPIM name into `Name` then. This way the implementation and interface should be cleaner.

Also please separate docs PR from feature PR because they seem to have different purpose.

korotovsky:
Thanks @nateGeorge.

Closing this as stale — the review feedback from 2025-11-26 (collapse the new fields into `Channel.Name`, and split docs from feature into two PRs) hasn't been addressed in about six months.

Happy to re-review if you reopen as two focused PRs.

PR #132, opened 2025-11-24
Per-request oauth token authentication in header

CLoui:


korotovsky:
Hi, thank you for the PR, this is very important and complicated topic you are trying to solve so I really appreciate.

Here are some things what needs to be fixed or fundamentally rethinked in my opinion, let me know you opinion on it as well:
- Slack MCP Server must stay not affiliated to any companies in the code, etc (i.e. LiteLLM)
- Not sure what was the reason for MemoryStore when token is always comes via header, consider the case if we run MCP server as lambda function,then it must be fully stateless or cache must be external (some redis-like resistance driver)
- Overall, I guess when we read token from header and try to disable cache then capabilities that were provided by cache must be reimplemented via extra requests to Slack API to maintain similar functionality.

PR #142, opened 2025-12-21
wait for isReady instead of erroring

ckhsponge:
Instead of erroring at startup if user and channel caches are not available, wait for them to load.

This could be configured with an env var if desired. And waiting 30s might be longer than needed.

### DIFF
--- a/pkg/handler/channels.go
+++ b/pkg/handler/channels.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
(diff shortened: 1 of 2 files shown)


korotovsky:
Closing — superseded by #277 (just merged), which fixes the underlying stdio-blocking root cause by removing the busy-wait loop and adding a `SkipCache()` provider method plus a `--no-cache` flag. The "wait for isReady instead of erroring" workaround is no longer needed. Thanks for surfacing the issue!

PR #149, opened 2026-01-01
feat: add Slack Lists API support

### LINKED ISSUE #95: Slack Lists support
Image

Hi 👋  

It would be very useful if `slack-mcp-server` could support Slack’s **Slack Lists** feature.  
Currently, I can query Slack Lists manually via HTTP requests, but native support in the MCP server would make this much easier and more consistent.  

## Key requirements
- Retrieve all Slack Lists in a workspace (with limit)
- Fetch all items within a list  
- Access item properties such as:  
  - whether it is completed or not  
  - completion timestamp  
  - assignee  
  - due date  
## Relevant Slack API endpoints
- : POST https://slack.com/api/slackLists.items.list 
- : POST  https://slack.com/api/slackLists.items.info

Would you consider adding support for these endpoints into `slack-mcp-server` so Lists can be queried directly (instead of raw HTTP requests)   ?

majiayu000:
Fixes #95

## Changes
- Add edge client methods for Slack Lists API (list, info, create, update, delete)
- Add MCP tool handlers for Lists operations
- Register 5 new MCP tools: lists_items_list, lists_items_info, lists_items_create, lists_items_update, lists_items_delete
- Support pagination via cursor for listing items

### DIFF
--- a/pkg/handler/lists.go
+++ b/pkg/handler/lists.go
@@ -1,0 +1,323 @@
+package handler
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/gocarina/gocsv"
+	"github.com/korotovsky/slack-mcp-server/pkg/provider"
+	"github.com/korotovsky/slack-mcp-server/pkg/provider/edge"
+	"github.com/mark3labs/mcp-go/mcp"
+	"go.uber.org/zap"
+)
+
+type ListItemCSV struct {
+	ID        string `json:"id" csv:"ID"`
+	ListID    string `json:"list_id" csv:"ListID"`
+	CreatedBy string `json:"created_by" csv:"CreatedBy"`
+	UpdatedBy string `json:"updated_by" csv:"UpdatedBy"`
+	Archived  bool   `json:"archived" csv:"Archived"`
+	ParentID  string `json:"parent_id" csv:"ParentID"`
+	Fields    string `json:"fields" csv:"Fields"`
+	Cursor    string `json:"cursor" csv:"Cursor"`
+}
+
+type ListsHandler struct {
+	apiProvider *provider.ApiProvider
+	logger      *zap.Logger
+}
+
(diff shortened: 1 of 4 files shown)


korotovsky:
Hi @majiayu000, have these edge API methods support all kind of tokens including `xoxb` & `xoxp` ?

majiayu000:
Hi @korotovsky, thanks for checking. I haven’t validated xoxb/xoxp yet. The Lists methods are routed via the Edge API client, which typically relies on browser session tokens (xoxc/xoxd). I can test OAuth tokens if needed and update the docs/implementation accordingly.

korotovsky:
This MCP supports three token types: browser (`xoxc` / `xoxd`), bot (`xoxb`), and OAuth2 (`xoxp`). In many cases, methods support both xoxb and xoxp on edge endpoints.

If an edge endpoint does not support a specific token, we should check whether the same functionality can be implemented using the REST API or another suitable API where that token is:
a) accepted, and
b) the required functionality exists.

If no such API exists, the tool must be disabled for that token type and the README updated (worst case).

That’s why, before adding new tools, I always ask to implement them for all token types. This gives users a wide range of features and makes the Slack MCP “so cool” 🙂

This is also one reason why many PRs are still open: adding new tools requires extra effort. I’d appreciate it if you could check and confirm that the new tools work with all three token types. If they don’t, please find and implement the missing APIs where possible.

korotovsky:
Hi @majiayu000, have you had a chance to validate other types?

korotovsky:
@majiayu000 ping here

majiayu000:
Apologies for the delay! I currently don't have the bandwidth to verify this with other token types (xoxb/xoxp) or implement the official APIs, as my workflow relies on the browser token (xoxc) which works with this edge implementation.

I understand if this implementation (using edge APIs) is not suitable for a general merge without broader support. Feel free to hold off on merging until someone (or myself in the future) can address the official API support.

majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

PR #155, opened 2026-01-20
fix: wait for cache to be ready before serving stdio requests

### LINKED ISSUE #135: slack-mcp-server shows as "connected" but Claude Code can't see any tools?
When I ask Claude Code it to verify access to slack using the mcp, it said: "the slack-mcp-server tools never loaded into my toolset - the server shows "Connected" but no mcp__slack__* tools appear."

I absolutely could have configured something wrong, but claude verified it could access the slack API directly using the tokens directly. So it's something else

Context:
- Claude Version: 2.0.61
- slack-mcp-server@1.1.26

Config:

```json
 "slack": {
	"type": "stdio",
	"command": "npx",
	"args": ["-y", "slack-mcp-server"],
	"env": {
		"SLACK_MCP_XOXP_TOKEN": "xoxp-[snip]",
		"SLACK_MCP_TEAM_ID": "[snip]",
		"SLACK_MCP_ADD_MESSAGE_TOOL": "true"
	}
},
```

Are there any obvious pointers on what I did wrong?

bhaskoro-muthohar:
## Summary

- Wait for cache to be ready before starting stdio server to fix race condition

## Problem

stdio transport starts accepting requests immediately while cache loading runs in a goroutine. For large workspaces (e.g., ~4.5k users, ~3.5k channels), the cache takes ~1 second to load from disk, but requests arrive in milliseconds - causing "cache not ready" errors.

SSE/HTTP transports don't have this issue because their server setup takes enough time for the cache to load.

## Solution

Add a simple wait loop before `ServeStdio()` that polls `IsReady()` until the cache is loaded.

## Test plan

- [x] Tested with large workspace (~4.5k users, ~3.5k channels)
- [x] Verified cache loads from disk in ~1 second
- [x] Verified requests succeed after fix

Fixes #135

### DIFF
--- a/cmd/slack-mcp-server/main.go
+++ b/cmd/slack-mcp-server/main.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/korotovsky/slack-mcp-server/pkg/provider"
 	"github.com/korotovsky/slack-mcp-server/pkg/server"
@@ -51,6 +52,12 @@ func main() {
 
 	switch transport {
 	case "stdio":
+		for {
+			if ready, _ := p.IsReady(); ready {
+				break
+			}
+			time.Sleep(100 * time.Millisecond)
+		}
 		if err := s.ServeStdio(); err != nil {
 			logger.Fatal("Server error",
 				zap.String("context", "console"),


korotovsky:
I think solution for `stdio` must be different - it feels that in `stdio` mode sync is "never" finished up the its end because Claude Code or other agent shuts it down and MCP server can not write consistent JSON on the disk. Instead, I think we must incrementally write cache to the disk, marking what was the last position so the next run actually "continues" sync from last interrupted page.

Would you be able to pick this up?

bhaskoro-muthohar:
Based on my investigation, the issue I encountered is specifically a race condition, not incomplete cache writes.

**What I observed:**
- Cache files exist and are complete on disk (~9MB users, ~775KB channels)
- Cache loading runs in a goroutine
- `ServeStdio()` starts immediately without waiting
- Requests arrive before the goroutine finishes loading cache from disk

**Evidence from logs:**
```
15:10:21 - "Caching users collection..."
15:10:21 - "Request received" → ERROR "cache not ready"
```

The cache was there, but the server accepted requests before loading it into memory.

Your concern about interrupted syncs and incremental writes is valid and would help first-run scenarios. But this PR addresses a different problem: existing valid cache not being loaded before stdio accepts requests.

Would you consider this as a fix for the race condition, with incremental cache writing as a separate enhancement?

korotovsky:
So basically it is in the `for` loop unless it does reads and deserialize JSON into go structs?

bhaskoro-muthohar:
Yes, exactly. The `for` loop blocks until `IsReady()` returns true, which happens after:

1. `RefreshUsers()` reads and deserializes `users_cache.json` into `map[string]slack.User`
2. `RefreshChannels()` reads and deserializes `channels_cache_v2.json` into `map[string]Channel`

Only then does `ServeStdio()` start accepting requests.

korotovsky:
Ok makes sense, thank you.

PR #156, opened 2026-01-20
feat(docker): Set mcp-server as ENTRYPOINT for improved usability

JAORMX:
## Summary

This change refactors the `Dockerfile` to use the `ENTRYPOINT` instruction, making the container easier and more intuitive to run with custom arguments.

### Problem

Previously, a container's entire command, including the `mcp-server` executable, was defined using `CMD`. This required users to re-specify the executable name whenever they wanted to override the default arguments.

For example, to change the transport mode with `docker run`, the command was unnecessarily verbose:

**Before:**
```bash
docker run <image> mcp-server --transport http
```

This is counter-intuitive and not aligned with common containerization practices.

### Solution

By setting `mcp-server` as the `ENTRYPOINT`, the `CMD` is now used only to provide default arguments (`--transport sse`). This allows users to override only the arguments when running the container, which is the standard and expected behavior.

**After:**
```bash
docker run <image> --transport http
```

This change improves the user experience and aligns the `Dockerfile` with containerization best practices, making it more flexible and easier to use in various environments.

### DIFF
--- a/docs/03-configuration-and-usage.md
+++ b/docs/03-configuration-and-usage.md
@@ -128,7 +128,6 @@ Open your `claude_desktop_config.json` and add the mcp server to the list of `mc
         "-e",
         "SLACK_MCP_XOXP_TOKEN",
         "ghcr.io/korotovsky/slack-mcp-server",
-        "mcp-server",
         "--transport",
         "stdio"
       ],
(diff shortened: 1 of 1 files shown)


korotovsky:
Hi @JAORMX, thanks for the PR, however all docs must be adjusted according to this change. https://github.com/korotovsky/slack-mcp-server/blob/master/docs/03-configuration-and-usage.md

JAORMX:
Sure! I'll update this tomorrow

JAORMX:
@korotovsky updated

PR #163, opened 2026-01-26
feat: Add conversations_unread tool to check for unread messages

cameronehrlich:
## Summary

This PR adds a new `conversations_unread` tool that lets users quickly check which channels and DMs have unread messages.

### The Problem

I wanted to use Claude Code to check if I had any unread Slack messages before diving into work. The existing MCP server has great tools for reading conversations and sending messages, but no way to see *which* conversations need attention.

### What I Tried

My first attempt used the `conversations.info` API to get unread counts for each channel individually. This worked but was **way too slow** - it made N API calls (one per channel), which caused timeouts when you have hundreds of channels.

### The Solution

I discovered the `client.counts` edge API that's already used elsewhere in this codebase. It returns unread status for **all channels in a single API call**. Much faster!

**Trade-off:** The `client.counts` API only returns `HasUnreads: true/false`, not the actual unread count. So the tool reports `UnreadCount: 1` for any channel with unreads. If exact counts are needed, that would require the slower per-channel approach.

## Changes

- **`pkg/provider/api.go`**: Added `ClientCounts` method to the `SlackAPI` interface and `MCPSlackClient`
- **`pkg/handler/conversations.go`**: Added `ConversationsUnreadHandler` and `UnreadChannel` struct
- **`pkg/server/server.go`**: Registered the new `conversations_unread` tool

## Usage

```
conversations_unread(channel_types: "im,mpim,public_channel,private_channel")
```

Returns CSV with columns: `ChannelID`, `ChannelName`, `UnreadCount`, `Purpose`

The `channel_types` parameter is optional and defaults to all types.

## Testing

- ✅ Tested locally with a user OAuth token (xoxp)
- ✅ Verified it returns results quickly (single API call)
- ✅ Verified empty response when no unreads (just CSV headers)
- ✅ Verified channel type filtering works correctly

## Screenshots

Example response when no unreads:
```
ChannelID,ChannelName,UnreadCount,Purpose
```

Example response with unreads:
```
ChannelID,ChannelName,UnreadCount,Purpose
D1234567890,@john.doe,1,DM with John Doe
C9876543210,#general,1,Company-wide announcements
```

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -51,6 +51,13 @@ type Message struct {
 	Cursor    string `json:"cursor"`
 }
 
+type UnreadChannel struct {
+	ChannelID   string `json:"channelID"`
+	ChannelName string `json:"channelName"`
+	UnreadCount int    `json:"unreadCount"`
+	Purpose     string `json:"purpose"`
+}
+
 type User struct {
 	UserID   string `json:"userID"`
 	UserName string `json:"userName"`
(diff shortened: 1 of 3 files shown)


korotovsky:
Hi @cameronehrlich, thank you for the PR, however it collides with almost [finalized PR](https://github.com/korotovsky/slack-mcp-server/pull/146) by @saoudrizwan. His PR covers much more cases and was nearly done on the review. I'd appreciate if you could join the conversation there, then we could move things a little faster.

PR #180, opened 2026-02-03
feat: add cache TTL and refresh-on-error for channel lookups

ljagiello:
## Summary

This PR addresses the issue where newly created Slack channels are not discovered because the cache is loaded once at startup and never refreshed from the Slack API.

### Problem

When MCP server starts, it loads the channel cache from a file. If a new channel is created in Slack after the cache was populated, the MCP server will not find it - leading to "channel not found" errors. This is particularly problematic after long server uptimes or restarts with stale cache files.

### Solution

Three complementary mechanisms:

1. **Cache TTL** - Configurable time-to-live for cache entries using file modification time
2. **Refresh-on-error** - When a channel lookup fails, force refresh the cache and retry once
3. **Rate limiting** - Prevent API abuse by limiting how often forced refreshes can occur

## Changes

- Add configurable cache TTL via `SLACK_MCP_CACHE_TTL` env var (default: 1h)
- Add configurable rate limiting via `SLACK_MCP_MIN_REFRESH_INTERVAL` env var (default: 30s)
- Use file modification time to detect cache staleness across restarts
- Add `ForceRefreshChannels()` and `ForceRefreshUsers()` methods for on-demand refresh
- Implement refresh-on-error pattern in `resolveChannelID()`: when a channel lookup fails, force refresh the cache and retry once
- Use atomic pointer swap (`sync/atomic.Pointer[T]`) for lock-free cache reads
- Add comprehensive test coverage for caching logic

### Configuration

**`SLACK_MCP_CACHE_TTL`** - How long before cache is considered stale:
- Duration strings: `"1h"`, `"30m"`, `"1h30m"`
- Numeric seconds: `"3600"`
- Zero to disable TTL (cache forever): `"0"`
- Negative values rejected, falls back to default

**`SLACK_MCP_MIN_REFRESH_INTERVAL`** - Minimum time between forced refreshes:
- Duration strings: `"30s"`, `"1m"`
- Numeric seconds: `"60"`
- Zero to disable rate limiting: `"0"`
- Negative values rejected, falls back to default

### Why file modification time?

The TTL check uses file modification time rather than in-memory timestamps. This correctly detects stale cache even after MCP server restarts - if the cache file is 3 days old when the server starts, it will be refreshed immediately.

### Why atomic pointers?

Cache reads use `atomic.Pointer[T]` for lock-free access. Writes build a new immutable snapshot and atomically swap the pointer. This eliminates lock contention on reads and avoids GC pressure from map copying.

## Test plan

- [x] Unit tests for `getCacheTTL()` parsing (7 cases including negative value rejection)
- [x] Unit tests for `getMinRefreshInterval()` parsing (7 cases)
- [x] Unit tests for cache expiry detection using file mtime
- [x] Unit tests for server restart with stale cache scenario
- [x] Unit tests for Channel struct JSON round-trip serialization
- [x] Unit tests for channel lookup by name (inverse map pattern)
- [x] Unit tests for channel ID format detection
- [x] Unit tests for refresh-on-error retry-once pattern
- [x] Unit tests for cache directory creation
- [x] All tests pass with `-race` flag (no data races)
- [x] Manual QA with real Slack workspace

### QA Results

**Cache expired (TTL=10s, cache age >10s):**
```json
{"level":"info","message":"Users cache expired, will refetch","cache_age":26969241617,"ttl":10000000000}
{"level":"info","message":"Channels cache expired, will refetch","cache_age":24149049945,"ttl":10000000000}
```

**Cache fresh (TTL=1h):**
```json
{"level":"info","message":"Loaded users from cache","count":37}
{"level":"info","message":"Loaded channels from cache and re-mapped DM names","count":134}
```

**TTL=0 disables expiry** - 6-year-old cache file still loads without refetch.

**Negative TTL rejected** - Falls back to default 1h.

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -177,7 +177,7 @@ func (ch *ConversationsHandler) ConversationsAddMessageHandler(ctx context.Conte
 		return nil, err
 	}
 
-	params, err := ch.parseParamsToolAddMessage(request)
+	params, err := ch.parseParamsToolAddMessage(ctx, request)
 	if err != nil {
 		ch.logger.Error("Failed to parse add-message params", zap.Error(err))
 		return nil, err
@@ -262,7 +262,7 @@ func (ch *ConversationsHandler) ReactionsAddHandler(ctx context.Context, request
 		return nil, err
 	}
 
-	params, err := ch.parseParamsToolReaction(request)
+	params, err := ch.parseParamsToolReaction(ctx, request)
 	if err != nil {
 		ch.logger.Error("Failed to parse add-reaction params", zap.Error(err))
 		return nil, err
@@ -298,7 +298,7 @@ func (ch *ConversationsHandler) ReactionsRemoveHandler(ctx context.Context, requ
 		return nil, err
 	}
 
-	params, err := ch.parseParamsToolReaction(request)
+	params, err := ch.parseParamsToolReaction(ctx, request)
 	if err != nil {
 		ch.logger.Error("Failed to parse remove-reaction params", zap.Error(err))
 		return nil, err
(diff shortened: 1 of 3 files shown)


ljagiello:
Hey @korotovsky, would you mind taking a look at this one?

PR #183, opened 2026-02-03
feat: add startup auth validation and workspace-isolated cache paths

Flare576:
## Summary
- Validates Slack tokens on startup via `AuthTest()` - server now fails fast with clear error message if tokens are invalid/expired
- Namespaces cache files by TeamID (e.g., `T12345_users_cache.json`) to prevent cross-workspace cache contamination when switching between Slack workspaces
- Logs authenticated workspace info on startup for debugging

## Problem
Previously, if you had multiple Slack workspaces configured (e.g., one for work, one for personal), they would share the same cache files. This caused several issues:
1. **Silent bad token**: If workspace A's cache existed but workspace B's tokens were used, the server would start "successfully" (cache loaded!) but API calls would fail
2. **Cache contamination**: Switching between workspaces would overwrite each other's user/channel caches

## Solution
1. **Auth validation on startup**: Call `AuthTest()` before loading cache. If auth fails, `logger.Fatal()` immediately with clear error message
2. **TeamID-prefixed caches**: Cache files are now namespaced by workspace: `T12345_users_cache.json`, `T12345_channels_cache_v2.json`

## Testing
- With valid tokens: Server starts, logs "Authenticated to Slack" with team/user info
- With invalid tokens: Server fails immediately with "Authentication failed - check your Slack tokens"
- Multiple workspaces: Each workspace gets its own cache files

## Depends On
- #179 (feat-users-search) - this branch includes those changes, please merge #179 first

PR #179, opened 2026-02-02
feat: add users_search tool for finding users by name/email

Flare576:
## Summary

Adds a `users_search` tool that searches for Slack users by name, email, or display name using the Edge API.

## Motivation / Use Case

Finding a specific user to interact with currently requires multiple API calls. For example, to answer "what did Quinn say in our DM?" you need to:
1. Get all DMs via `channels_list`
2. Iterate through users to find Quinn's user ID (no search API available)
3. Match user ID to DM channel
4. Finally call `conversations_history`

This can take 8+ API calls for a simple lookup.

**With `users_search`, I can now do workflows like:**
1. Search for "Coworker A" → get their UserID and DM channel
2. Get the latest message from their DM
3. React to that message with `:eyes:` emoji
4. Look up information with another tool
5. Search for "Coworker B" → get their UserID
6. Send a message to Coworker B with the information
7. Remove `:eyes:` emoji from Coworker A's message
8. React to Coworker A's message with `:thumbsup:`

## Implementation

- Adds `UsersSearch` method to the Edge client (`pkg/provider/edge/users.go`)
- Calls `https://edgeapi.slack.com/cache/{team}/users/search` endpoint
- Returns CSV with: UserID, UserName, RealName, DisplayName, Email, Title, DMChannelID

## Token Compatibility

| Token Type | Supported |
|------------|-----------|
| xoxc/xoxd (browser) | ✅ |
| xoxp (user OAuth) | 🟡 |
| xoxb (bot) | 🟡 |

This follows the same pattern as `conversations_unreads` - the Edge API only accepts browser session tokens.

## Testing

Tested with browser tokens - successfully searches and returns user data:
```
UserID,UserName,RealName,DisplayName,Email,Title,DMChannelID
U01XXXXXXXXX,jane.doe,Jane Doe,jane.doe,jane.doe@example.com,Senior Engineer,
```

PR #184, opened 2026-02-03
feat: add Enterprise Grid channel name resolution fallback

Flare576:
## Summary
- Adds fallback resolution for `#channel-name` syntax when the `SearchChannels` API fails with `team_is_restricted` error (common in Enterprise Grid)
- Falls back to `search.messages` API with `in:#channel-name` query to extract channel ID

## Problem
In Slack Enterprise Grid environments with multiple workspaces, the `search.modules.channels` API (used by `SearchChannels`) returns `team_is_restricted` even when the user has access to the channel. This prevents using `#channel-name` syntax for message operations.

However, the `search.messages` API handles channel names server-side and works correctly - Slack resolves `in:#channel-name` internally before searching.

## Solution
1. **Primary path**: Try `SearchChannels` API (works in most environments)
2. **Fallback path**: If `team_is_restricted` error, use `search.messages` with `in:#channel-name` query and extract channel ID from results
3. **Cache update**: On successful resolution, update the in-memory cache for future lookups

## Changes
- `pkg/provider/api.go`: Add `SearchChannels` method to `SlackAPI` interface and `MCPSlackClient`
- `pkg/handler/conversations.go`:
  - `resolveChannelIDWithFallback`: Main resolution with Enterprise Grid fallback
  - `resolveUserDMChannelWithFallback`: DM channel resolution with user search fallback
  - `resolveChannelViaSearch`: Extract channel ID from search results
  - `isTeamRestrictedError`: Detect the Enterprise Grid restriction error
  - `paramFormatChannel/paramFormatUser`: Updated to use fallback methods

## Testing
- Tested in ASU Enterprise Grid environment where `SearchChannels` consistently returns `team_is_restricted`
- `#channel-name` syntax now resolves correctly via the search fallback
- Standard Slack workspaces continue to use the primary `SearchChannels` path

## Depends On
- #183 (startup auth validation) - this PR builds on those changes
- #179 (users_search) - for UsersSearch method

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -625,52 +625,7 @@ func isChannelAllowed(channel string) bool {
 }
 
 func (ch *ConversationsHandler) resolveChannelID(ctx context.Context, channel string) (string, error) {
-	if !strings.HasPrefix(channel, "#") && !strings.HasPrefix(channel, "@") {
-		return channel, nil
-	}
-
-	// First attempt: try to resolve from current cache
-	channelsMaps := ch.apiProvider.ProvideChannelsMaps()
-	chn, ok := channelsMaps.ChannelsInv[channel]
-	if ok {
-		return channelsMaps.Channels[chn].ID, nil
-	}
-
-	// Channel not found - try refreshing cache and retry once
-	ch.logger.Debug("Channel not found in cache, attempting refresh",
-		zap.String("channel", channel))
-
-	refreshErr := ch.apiProvider.ForceRefreshChannels(ctx)
-	wasRateLimited := errors.Is(refreshErr, provider.ErrRefreshRateLimited)
-
-	if refreshErr != nil && !wasRateLimited {
-		ch.logger.Error("Failed to refresh channels cache",
-			zap.String("channel", channel),
-			zap.Error(refreshErr))
-		return "", fmt.Errorf("channel %q not found and cache refresh failed: %w", channel, refreshErr)
-	}
-
-	// If rate-limited, cache wasn't refreshed - no point in a second lookup
-	if wasRateLimited {
-		ch.logger.Warn("Channel not found; cache refresh was rate-limited",
-			zap.String("channel", channel))
-		return "", fmt.Errorf("channel %q not found (cache refresh was rate-limited, try again later)", channel)
-	}
-
-	// Second attempt after successful refresh
-	channelsMaps = ch.apiProvider.ProvideChannelsMaps()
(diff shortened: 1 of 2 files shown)


korotovsky:
@Flare576 pls rebase with master

korotovsky:
To be honest, the entire implementation looks a bit off. In my opinion, the differences between the normal and fallback behavior should be handled inside the Slack provider. Once this logic goes beyond the provider API, it becomes a leaked abstraction, which isn’t ideal.

I don’t think it makes sense to merge this at this stage. We can either keep it open to gather feedback from the community and explore a better approach, or close it for now so you can come back later with an improved solution.

Flare576:
Thanks for taking the time to review and provide feedback on the original implementation. You were right that the fallback logic was leaking into the handler layer.

I've reworked this to encapsulate the Enterprise Grid channel resolution entirely within `ApiProvider`:

- Added `ResolveChannelByName(ctx, channelName)` to the provider, which handles the full resolution strategy (cache → refresh → SearchChannels fallback for Enterprise Grid with xoxc/xoxd)
- The handler's `resolveChannelID` is now just a one-liner delegating to the provider
- Channels found via the SearchChannels fallback get added to the cache for future lookups

Is this the kind of isolation you had in mind, or would you prefer a different approach?

korotovsky:
Yeah I think it's better now, I think necessity of `resolveChannelID` in itself kind of is void now, we could have directly call provider function?

Let's keep this open for some time for peer community review.

korotovsky:
Thanks @Flare576 — and apologies for not getting to this sooner.

Closing as superseded. The Enterprise Grid `SearchChannels` fallback you proposed here has already landed on master via a separate channel-resolution refactor — see `ResolveChannelByName` in `pkg/provider/api.go` (around line 1398). The flow there is:

1. Look in the local cache
2. If miss, call `ForceRefreshChannels` (rate-limit-aware)
3. If still miss **and** Enterprise Grid + non-OAuth token, fall back to the `SearchChannels` edge API — matching your `team_is_restricted`-style use case
4. On hit, add the resolved channel to the cache via `addChannelToCache`

That covers the same gap your PR was addressing, with the cache-refresh-and-add-to-cache steps integrated. Since the functionality is now in master, there's no good way to merge this PR without large amounts of duplicate work.

Really appreciate the diagnosis and the original fix — it informed the design of the version that landed. If you've found cases where `ResolveChannelByName` still doesn't handle your Enterprise Grid setup correctly, please open an issue and we'll address it directly.

PR #188, opened 2026-02-05
Include Fields and TitleLinks from attachments

brokensandals:
This includes additional information from `Attachment` objects in the text returned by the MCP.

One motivation for this is that Honeycomb's Slack app generates some messages where the only useful information is in the `Field`s of an `Attachment`.

Disclosure: Claude generated this code and I'm not experienced with Go. But the code looks reasonable to me, `make clean && make test` passes, and I built the docker image locally and tested the MCP with Claude Code and it seems to work as desired.

### DIFF
--- a/pkg/text/text_processor.go
+++ b/pkg/text/text_processor.go
@@ -18,7 +18,11 @@ func AttachmentToText(att slack.Attachment) string {
 	var parts []string
 
 	if att.Title != "" {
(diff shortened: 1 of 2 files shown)


korotovsky:
Closing in favor of #272, which covers the same attachment `Fields` extraction with stronger motivation (specifically targeting bot/integration messages where structured content was silently dropped) and includes tests. Thanks for the work — please review #272 if you'd like to add review comments there.

PR #193, opened 2026-02-08
feat: add user group management tools

RedSlowpoke:
## Summary

Add five new tools for managing Slack user groups (subteams):

- **`usergroups_list`** - List all user groups in the workspace with optional filters
- **`usergroups_me`** - Manage your own membership: list groups you're in, join, or leave
- **`usergroups_create`** - Create a new user group with name, handle, and description
- **`usergroups_update`** - Update group metadata (name, handle, description, default channels)
- **`usergroups_users_update`** - Replace all members of a group

### Key Features

The `usergroups_me` tool provides a convenient way to join or leave groups without needing to know the full member list. It handles fetching current members and updating the list automatically - addressing the limitation that Slack's API only supports replacing the entire member list.

### Required OAuth Scopes

- `usergroups:read` - For list operations
- `usergroups:write` - For create/update/join/leave operations

### Tool Descriptions

All tools include detailed descriptions optimized for LLM understanding, explaining:
- What user groups are (mention groups like @engineering)
- When to use each tool
- Output format (CSV vs JSON)
- ID format and how to get IDs
- Warnings about destructive operations

## Test Plan

- [x] Build passes: `go build ./...`
- [x] Vet passes: `go vet ./...`
- [x] Manual testing with MCP Inspector
- [x] Verified idempotent join/leave (returns message if already member/not member)


korotovsky:
Bringing new tools into the MCP is one of the most complicated things, because I try to keep them functional on the same level among all three types of tokens: `xoxp`, `xoxb` and `xoxd`/`xoxc`. Have you researched possible solutions how usergroups could be managed via `xoxc`/`xoxd`?

RedSlowpoke:
I've manually tested with xoxc/xoxd and it works just fine.

RedSlowpoke:
Additionaly, I want to add, that some of the tools that I propose in this PR are desctructive.
I think it may be the best to finalize the https://github.com/korotovsky/slack-mcp-server/pull/173 before finalizing this one.

korotovsky:
I agree, was about to suggest the same. But please update OAuth JSON spec in docs for completeness anyway.

RedSlowpoke:
I've tested with xoxb as well and it make it possible for list and even _me works to join and leave for the bot.
So I've tested with all 3 types of tokens manually and it seems working for me on my corporate slack.

korotovsky:
Thank you!

PR #194, opened 2026-02-08
fix: rename payload parameter to text in conversations_add_message

### LINKED ISSUE #181: conversations_add_message: parameter name mismatch between schema and implementation
## Bug Description

  The `conversations_add_message` tool has a mismatch between the expected parameter name in the schema/documentation and
  the actual implementation.

  ## Expected Behavior

  Based on the tool description and error message, the parameter should be named `text`:
  ```json
  {"channel_id": "C0ABZLSU86N", "text": "hello"}
  ```

  ## Actual Behavior

  The implementation expects `payload` instead of `text`:
  ```go
  // pkg/handler/conversations.go line ~1023
  msgText := request.GetString("payload", "")
  if msgText == "" {
      return nil, errors.New("text must be a string")  // misleading error message
  }
  ```

  Using `text` returns error:
  ```
  MCP error -32603: text must be a string
  ```

  Using `payload` works correctly:
  ```json
  {"channel_id": "C0ABZLSU86N", "payload": "hello"}
  ```

  ## Environment

  - slack-mcp-server version: v1.1.28
  - MCP client: mcporter 0.7.3

  ## Suggested Fix

  Either:
  1. Change the implementation to accept `text` parameter (and keep `payload` for backward compatibility)
  2. Update the tool schema and error message to say `payload`

  Option 1 is preferred as `text` is more intuitive for a message posting tool.

RedSlowpoke:
## Summary
- Renamed `payload` parameter to `text` in `conversations_add_message` tool for better intuitiveness
- Added backward compatibility - existing integrations using `payload` will continue to work
- Error message now matches the expected parameter name

Fixes #181


korotovsky:
~Could you please clean up PR from usergroups code?~

PR #196, opened 2026-02-09
fix: return tool errors as isError results instead of JSON-RPC errors

aminsaedi:
Per the MCP specification, tool execution errors should be returned as tool results with isError=true, not as JSON-RPC protocol errors (-32603). This allows LLMs to see the error message and self-correct by retrying with different parameters.

Previously, all handler errors were returned as Go errors which mcp-go converted into JSON-RPC -32603 responses. MCP clients either crash or silently discard these, so the LLM never sees what went wrong.

This adds a middleware that catches handler errors and converts them to mcp.NewToolResultError() responses. The middleware is registered before the logger and auth middlewares, so it catches all tool handler errors regardless of which tool triggered them.

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -1,10 +1,20 @@
 package server
 
 import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
 	"os"
 	"testing"
 
+	"github.com/mark3labs/mcp-go/client"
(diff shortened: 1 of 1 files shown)


korotovsky:
Could you please add integration test for this?

aminsaedi:
@korotovsky pushed

PR #199, opened 2026-02-11
feat: Add xoxp/xoxb token support for conversations_unreads

niebloomj:
## Summary

Adds fallback implementation for `conversations_unreads` tool that works with xoxp and xoxb tokens.

## ⚠️ Stacked PR

**This PR is stacked on top of #171** and includes those commits. It should be reviewed after #171 is merged, or reviewed with the understanding that it builds upon that PR's functionality.

## Problem

The current implementation of `conversations_unreads` (from #171) uses the `ClientCounts` API, which only works with browser session tokens (xoxc/xoxd). When used with OAuth tokens (xoxp/xoxb), it executes without error but returns empty results.

## Solution

This PR adds an automatic fallback that:
1. Attempts to use `ClientCounts` first (fast path for xoxc/xoxd)
2. Detects when `ClientCounts` returns empty results  
3. Falls back to iterating through channels using `conversations.info` to check `unread_count` on each channel

This approach was validated in a Python implementation before porting to Go.

## Implementation Details

- **New method**: `getUnreadsUsingConversationsInfo` implements the per-channel approach
- **Auto-detection**: Checks if `ClientCounts` returns zero channels and switches to fallback
- **API addition**: Added `GetConversationInfoContext` to `SlackAPI` interface
- **Logging**: Info message logged when using fallback method
- **Compatibility**: Maintains same output format and filtering options as original

## Limitations with xoxp/xoxb tokens

- The `mentions_only` filter is not supported (no access to mention counts via `conversations.info`)
  - A warning is logged if `mentions_only` is requested
- `lastRead` and `latest` timestamps are not available via `conversations.info` (returned as empty strings)
- Slower than `ClientCounts` approach as it requires one API call per channel

## Testing

- ✅ Code compiles successfully
- ✅ Based on working Python implementation using the same approach

## Related

- Builds on #171 by @Flare576
- Addresses the xoxp/xoxb token limitation mentioned in #171

🤖 Generated with Claude Code

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -647,6 +647,14 @@ func (ch *ConversationsHandler) ConversationsUnreadsHandler(ctx context.Context,
 		zap.Int("mpims", len(counts.MPIMs)),
 		zap.Int("ims", len(counts.IMs)))
 
+	// Check if ClientCounts returned empty results (happens with xoxp/xoxb tokens)
+	// If so, fall back to using conversations.info per-channel approach
+	totalCounts := len(counts.Channels) + len(counts.MPIMs) + len(counts.IMs)
+	if totalCounts == 0 {
+		ch.logger.Info("ClientCounts returned empty results, falling back to xoxp-compatible approach using conversations.info")
+		return ch.getUnreadsUsingConversationsInfo(ctx, params)
+	}
+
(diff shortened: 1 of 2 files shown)


niebloomj:
Approach was merged into https://github.com/korotovsky/slack-mcp-server/pull/171 so I am closing this

PR #202, opened 2026-02-13
fix: token-type routing and validation for conversations_unreads

140am:
## Summary

Addresses the 3 review comments from @korotovsky on #171, plus fixes a silent bug in `GetMutedChannels` for xoxp tokens. Built on top of #171 (`3e0e4ea`).

- Pre-check token type in handler using `IsOAuth()`/`IsBotToken()` instead of calling `ClientCounts` speculatively and catching `not_allowed_token_type` via string matching
- Gate `conversations_unreads` registration behind `!provider.IsBotToken()` in `server.go` — bot tokens don't support unread tracking
- Cap xoxp fallback scan depth to ~300 API calls max (down from ~2000+ unbounded) and surface scan metadata + limitations in the response
- Fix `GetMutedChannels` silently swallowing `ok: false` responses from `users.prefs.get` — the `read` scope it requires is a [legacy-only scope](https://docs.slack.dev/legacy/legacy-custom-integrations/legacy-custom-integrations-tokens) not available in OAuth v2, so xoxp tokens always fail. Without `validate()`, the error was swallowed and muted channels leaked into results
- Add `shouldAddTool()` gating and `ValidToolNames` constants for both `conversations_unreads` and `conversations_mark` to match the established pattern

## Commits

1. **`b34dafa`** — Pre-check token type to avoid faulty `ClientCounts` call
2. **`58c46c8`** — Exclude `conversations_unreads` tool for bot tokens
3. **`c3c0daf`** — Cap xoxp fallback scan depth and document limitations
4. **`0d216e8`** — Validate `GetMutedChannels` response to surface xoxp failures
5. **`6c71838`** — Add `shouldAddTool` gating for `conversations_unreads` and `conversations_mark`

## Token routing (after)

| Token | Path | API calls |
|-------|------|-----------|
| xoxc/xoxd | `ClientCounts` (fast path) | 1 |
| xoxp | `getUnreadsViaConversationsInfo` (per-channel scan, capped) | ~300 max |
| xoxb | Tool not registered | n/a |

## Testing

Verified with real tokens on a workspace with ~2,700 channels and 129 muted channels:
- xoxc: 1 API call, all 129 muted channels filtered correctly
- xoxp: ~292 API calls (~30s), response includes muted filtering warning and scan metadata
- Build passes, existing tests updated

## Note

This PR includes the base commits from #171. It can either be merged independently or rebased on top of #171 once that lands.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -32,6 +32,8 @@ const (
 	ToolReactionsRemove             = "reactions_remove"
 	ToolAttachmentGetData           = "attachment_get_data"
 	ToolConversationsSearchMessages = "conversations_search_messages"
+	ToolConversationsUnreads        = "conversations_unreads"
+	ToolConversationsMark           = "conversations_mark"
 	ToolChannelsList                = "channels_list"
 	ToolUsergroupsList              = "usergroups_list"
 	ToolUsergroupsMe                = "usergroups_me"
@@ -48,6 +50,8 @@ var ValidToolNames = []string{
 	ToolReactionsRemove,
 	ToolAttachmentGetData,
 	ToolConversationsSearchMessages,
+	ToolConversationsUnreads,
+	ToolConversationsMark,
 	ToolChannelsList,
 	ToolUsergroupsList,
 	ToolUsergroupsMe,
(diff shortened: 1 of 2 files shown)


140am:
Closing — opened against wrong base. These fixes should go into Flare576's feature branch so they fold into #171.

Issue #153, opened 2026-01-13
MPIM channels not returned by channels_list despite being requested


dguido:
## Summary

The `channels_list` tool returns 0 MPIM (group DM) channels even when:
1. `channel_types` includes `mpim`
2. The token has `mpim:read`, `mpim:write`, `mpim:history` scopes
3. The user is a member of multiple MPIMs

## Investigation

I traced through the code and confirmed the MCP server is correctly requesting mpim channels:

- `api.go:25` - `AllChanTypes` includes `"mpim"`
- `api.go:580` - `RefreshChannels` calls `GetChannels(ctx, AllChanTypes)`
- `api.go:700-703` - Iterates over all channel types including mpim
- `api.go:636-641` - Calls Slack API with `Types: []string{channelType}` for each type

The cache file confirms the issue:
```bash
$ jq 'keys | length' ~/Library/Caches/slack-mcp-server/channels_cache_v2.json
725

$ jq 'to_entries | map(select(.value.mpim == true)) | length' ~/Library/Caches/slack-mcp-server/channels_cache_v2.json
0
```

725 channels cached, 0 are mpim.

## Workaround

Users must obtain MPIM channel IDs externally (e.g., Slack's "Copy link" feature) and provide them directly to `conversations_history` or `conversations_add_message`.

Search can find messages in MPIMs but displays the friendly name (`#mpdm-user1--user2--user3-1`) rather than the raw channel ID (`G6F8RH02G`).

## Environment

- Token type: User OAuth (`xoxp-...`)
- Scopes include: `mpim:read`, `mpim:write`, `mpim:history`, `groups:read`, `groups:history`
- MCP server version: latest via `npx slack-mcp-server@latest`

## Possible causes

1. Slack's `conversations.list` API may have undocumented restrictions on mpim enumeration
2. OAuth tokens may behave differently than session tokens for mpim listing
3. There may be a pagination or rate limiting issue specific to mpim channels

## Request

Could you investigate whether this is:
- A Slack API limitation that should be documented
- A bug in how the API response is processed
- Something that works with `xoxc`/`xoxd` tokens but not OAuth tokens

Issue #148, opened 2025-12-31
channels_list returns incomplete results (only 6 of 15+ accessible channels)


t4dhg:
## Description

The `channels_list` tool returns only 6 channels, despite the user having access to 15+ channels (confirmed via `conversations_search_messages` which finds messages across all of them).

## Steps to Reproduce

1. Configure MCP with a user token (`xoxp-...`) that has scopes: `channels:read`, `channels:history`, `groups:read`, `groups:history`, `search:read`, etc.
2. Call `channels_list` with `channel_types=public_channel,private_channel,im,mpim` and `limit=999`
3. Only 6 channels are returned

## Expected Behavior

All channels the user is a member of should be listed (15+ in this case).

## Actual Behavior

Only 6 channels returned (out of 15+ accessible).

## Evidence That Token Has Broader Access

Using `conversations_search_messages`, messages are found in many more channels than `channels_list` returns. This confirms the token has access to those channels, but `channels_list` doesn't return them.

## Environment

- Package: `slack-mcp-server@latest` via npx
- Transport: stdio
- Token type: User OAuth Token (`xoxp-...`)
- User Token Scopes: `admin`, `channels:history`, `channels:read`, `groups:history`, `groups:read`, `mpim:history`, `search:read`, `search:read.private`, `groups:write.topic`

## Possible Cause

The Slack API's `conversations.list` requires explicit `types` parameter and handles pagination. The issue might be:
1. Not all channel types being requested
2. Pagination not being followed
3. Some filtering logic excluding channels

Thanks for the great MCP server!

buildgreatthings:
I have a similar issue. When I access the API directly with my token, it's fine.

TranHung92:
+1 this issue is blocking me from using this slack mcp.

korotovsky:
Do you guys all have 6 channels returned despite of large amount of channels in JSON cache file?

buildgreatthings:
My cache shows public channels, direct messages, an direct bot messages. I am missing private channels. I can confirm my API key has access.

korotovsky:
What MCP request was made? Client/etc? Was it with default/no parameters or specified to some values?

buildgreatthings:
```{
  "method": "tools/call",
  "params": {
    "name": "channels_list",
    "arguments": {
      "channel_types": "private_channel",
      "limit": 100
    },
    "_meta": {
      "progressToken": 1
    }
  }
}
```

XOXP token

[...]

PR #218, opened 2026-03-03
fix: channels_list returns empty results on Enterprise Grid with xoxc/xoxd tokens

### LINKED ISSUE #120: MCP able to pull users info but not channels
We have a large enterprise that has multiple workspaces (in case that matters).

The MCP is able to list users and I can see them in the cache json file. When it tries to pull channels however, it appears to have received an empty result. The cache file for channels only contains `null`.

I'm using this MCP through Goose.

Prompt is:
> What Slack channels are available with "test" in the name?

Output from Goose is:
```
Channels List channel_types, limit

Tool Details
channel_types
public_channel,private_channel,im
limit
100

Output
ID,Name,Topic,Purpose,MemberCount,Cursor

11:22 AM
There are no Slack channels available with "test" in the name. If you need to search for a different variation or review the full list of available channels, let me know!
```

I'm not able to find where Goose might be putting any additional logging for the MCPs. Paths that I'm seeing cited by their documentation don't seem to exist.

georgebashi:
# fix: channels_list returns empty results on Enterprise Grid with xoxc/xoxd tokens

Fixes #120. Related: #148, #153.

## Problem

On Enterprise Grid workspaces using `xoxc`/`xoxd` session tokens (non-OAuth), `channels_list` returns only CSV headers with no data rows. The channel cache file contains `null`.

Three bugs combine to cause this:

1. **Edge API discards all channels on supplementary call failure.** `GetConversationsContext` in the edge client runs 3 concurrent goroutines (ClientUserBoot, IMList, SearchChannels) to fetch channels, then calls `ClientCounts` to discover MPIMs. On restricted teams, `ClientCounts` returns `team_is_restricted` — and the old code discarded *all* successfully-fetched channels. The same discard-on-error pattern existed in the goroutine result collector: if any one goroutine failed, channels from the other two were thrown away.

2. **No fallback from edge API to standard API.** For Enterprise Grid + non-OAuth tokens, the server exclusively used the edge API. If it failed, the error was propagated with no channels — even though the standard `conversations.list` API also works with `xoxc` tokens in many Enterprise Grid setups.

3. **Null cache poisoning.** When `GetChannels` returned nil (due to bugs 1+2), `json.MarshalIndent(nil)` produced `"null"`, which was written to the cache file. On subsequent startups, `json.Unmarshal("null")` into a slice *succeeds with no error* (a Go quirk), producing an empty slice that was treated as valid cached data — permanently serving empty results until the cache TTL expired.

Additionally, `GetChannels` called `GetConversationsContext` 4 times (once per channel type), but the edge API ignores the type filter and returns everything in one shot. The standard `conversations.list` API also supports multiple types per request. This made every cache refresh do 4× the necessary API calls.

### `pkg/provider/edge/slacker.go`

- **Goroutine errors are now non-fatal.** Channels from successful sources are kept; only if *every* source fails is the error propagated. This means if `SearchChannels` times out but `ClientUserBoot` and `IMList` succeed, you still get channels + IMs.
- **`ClientCounts` and `ConversationsGenericInfo` errors are non-fatal.** MPIM discovery is supplementary — losing MPIMs is better than losing all channels.

### `pkg/provider/api.go`

- **Edge → standard API fallback.** When the edge API fails, fall back to `conversations.list`. An `edgeFailed` flag remembers the failure so subsequent pagination calls go straight to the standard API instead of retrying edge on every page.
- **Null cache guards (defense-in-depth).** Don't write cache when channels slice is empty. Treat empty/null unmarshaled cache as invalid (trigger refetch). Same guard for users cache for consistency.
- **Single-call channel fetch.** Extract `getChannelsMultiType` that accepts all types at once. `GetChannels` now makes one API call instead of four. `GetChannelsType` is preserved as a thin wrapper.

## Testing

31 new unit tests across 4 test files, covering:
- Partial result collection with various failure combinations (`slacker_test.go`)
- Edge/standard routing decision matrix and `edgeFailed` flag behavior (`api_edge_fallback_test.go`)
- `json.Unmarshal("null")` behavior, empty cache detection, write guards (`api_cache_test.go`)
- Channel type filtering logic and `AllChanTypes` constant (`api_channels_test.go`)

```
$ go test ./pkg/provider/... ./pkg/provider/edge/...
ok  	github.com/korotovsky/slack-mcp-server/pkg/provider	0.829s
ok  	github.com/korotovsky/slack-mcp-server/pkg/provider/edge	0.381s
```

Manually verified on a large Enterprise Grid workspace (~32K channels, ~11K users) with `xoxc`/`xoxd` tokens. All channel types (public, private, IM, MPIM) return data correctly.

### DIFF
--- a/pkg/provider/api.go
+++ b/pkg/provider/api.go
@@ -1006,8 +1006,12 @@ func (ap *ApiProvider) GetSlackConnect(ctx context.Context) ([]slack.User, error
 }
 
 func (ap *ApiProvider) GetChannelsType(ctx context.Context, channelType string) []Channel {
+	return ap.getChannelsMultiType(ctx, []string{channelType})
+}
+
+func (ap *ApiProvider) getChannelsMultiType(ctx context.Context, channelTypes []string) []Channel {
 	params := &slack.GetConversationsParameters{
-		Types:           []string{channelType},
+		Types:           channelTypes,
 		Limit:           999,
 		ExcludeArchived: true,
 	}
@@ -1027,8 +1031,8 @@ func (ap *ApiProvider) GetChannelsType(ctx context.Context, channelType string) 
 		}
 
 		channels, nextcur, err = ap.client.GetConversationsContext(ctx, params)
-		ap.logger.Debug("Fetched channels for ",
-			zap.String("channelType", channelType),
+		ap.logger.Debug("Fetched channels",
+			zap.Strings("channelTypes", channelTypes),
 			zap.Int("count", len(channels)),
 		)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


korotovsky (approved):
We verified this PR on another large Enterprise Grid Slack. Everything works.

Issue #222, opened 2026-03-04
Resolve user mentions in message text to display names


gJigsaw:
## Problem

When reading messages via `conversations_history`, `conversations_search_messages`, or `conversations_replies`, user mentions embedded in message text remain as raw Slack IDs (e.g., `<@U0AFE8TJQ0L>`) instead of being resolved to human-readable names.

Example output from `conversations_history`:

```
Text
"U0A0FUA8ZRQ U09DFRLDJ7Q U02UDJPB78T U02FVMBKEP7 STACKKKED lineup today"
```

The message author gets resolved (via `UserName` and `RealName` columns), but mentions within the text don't.

## Expected behavior

Mentions in message text should be resolved to names:

```
"@Katherine @Kyle @Tom @Sarah STACKKKED lineup today"
```

## Current workaround

Call `users_search` for each user ID found in message text, then manually substitute. This is tedious and error-prone for consumers.

## Suggested implementation

The server already has user data cached and resolves `msg.User` to names via `getUserInfo()`. The same lookup could be applied to `<@USERID>` patterns in `msg.Text` before returning results.

In `pkg/handler/conversations.go`, after this line:
```go
Text: text.ProcessText(msgText),
```

Add a step to resolve `<@USERID>` patterns using the existing `usersMap`.

psyberck:
Hey, can I work on this issue as my first contribution to this cool project? This issue looks like a low hanging fruit I can quickly apply a solution. Thanks.

korotovsky:
Hi @ZihaoXue1995, yes, feel free to pick it up!

PR #233, opened 2026-03-10
refactor: resolves user mentions in message text to display names

psyberck:
# Summary (Close Issue #222 )
Reading messages via `conversations_history`, `conversations_search_messages`, or `conversations_replies`, user mentions embedded in message text will be displayed as usernames.

# Implementations
- `ProcessText` now delegates entirely to `filterSpecialChars` instead of having its own mention-handling logic
- `filterSpecialChars` gained a `userMaps map[string]slack.User` parameter — user mention resolution (<@U...> → @name) happens in a single pass alongside link/special-char processing

## Pros
  - Single pass: mention resolution and special-char filtering happen together — fewer string iterations

## Cons
  - ~~`filterSpecialChars` is no longer pure: it now takes external state (userMaps), making it harder to test in isolation from user data~~
  - `filterSpecialChars` remains untouched

# Test plan
- [x]  go build ./... — compiles cleanly
- [x]  go vet ./... — no warnings
- [x]  TestFilterSpecialCharsWithCommas — updated to add 9 new test cases for user mentions
- [x]  All existing unit tests still pass
- [x]  Manual testing with xoxc/xoxd tokens
- [x]  Manual testing with xoxp/xoxb tokens

Let me know what you think, thanks!

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -1599,7 +1599,7 @@ func (ch *ConversationsHandler) convertMessagesFromHistory(ctx context.Context, 
 			UserID:        msg.User,
 			UserName:      userName,
 			RealName:      realName,
-			Text:          processText(msgText, usersMap.Users),
+			Text:          processText(msgText, resolver.usersMap.Users),
 			Channel:       channel,
 			ThreadTs:      msg.ThreadTimestamp,
 			Time:          timestamp,
@@ -1659,7 +1659,7 @@ func (ch *ConversationsHandler) convertMessagesFromSearch(ctx context.Context, s
 			UserID:    msg.User,
 			UserName:  userName,
 			RealName:  realName,
-			Text:      processText(msgText, usersMap.Users),
+			Text:      processText(msgText, resolver.usersMap.Users),
 			Channel:   fmt.Sprintf("%s (#%s)", msg.Channel.ID, msg.Channel.Name),
 			ThreadTs:  threadTs,
 			Time:      timestamp,
@@ -2461,18 +2461,10 @@ func hasImageBlocks(blocks slack.Blocks) bool {
 
 func processText(s string, userMaps map[string]slack.User) string {
 	protected := s
-	userMentions := text.UserMentionRegex.FindAllString(s, -1)
-	for i, userMention := range userMentions {
-		placeholder := "___USER_AT_PLACEHOLDER___" + string(rune(48+i)) + "___"
-		protected = strings.Replace(protected, userMention, placeholder, 1)
-	}
-
-	cleaned := text.ProcessText(protected)
-	// Restore the @ sign
-	for i, userMention := range userMentions {
+	matches := text.UserMentionRegex.FindAllStringSubmatch(protected, -1)
+	for _, match := range matches {
+		userId := match[1]
 		var userName string
-		placeholder := "___USER_AT_PLACEHOLDER___" + string(rune(48+i)) + "___"
-		userId := userMention[2 : len(userMention)-1]
 		if u, ok := userMaps[userId]; ok {
 			name := u.Profile.DisplayName
 			if name == "" {
@@ -2485,7 +2477,8 @@ func processText(s string, userMaps map[string]slack.User) string {
 		} else {
 			userName = userId
 		}
-		cleaned = strings.Replace(cleaned, placeholder, fmt.Sprintf("@%s", userName), 1)
+		protected = strings.Replace(protected, match[0], "@"+userName, 1)
 	}
+	cleaned := text.ProcessText(protected)
 	return cleaned
 }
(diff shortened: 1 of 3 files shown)


chriscoey (commented):
Nice work — the placeholder approach is clean, follows the existing URL protection pattern, and the test coverage is solid. A few suggestions:

**1. Display name vs username**

`u.Name` is the Slack *username* (e.g., `chris.coey`), but what users see in the UI is `Profile.DisplayName` or `RealName`. Consider:
```go
name := u.Profile.DisplayName
if name == "" {
    name = u.RealName
}
if name == "" {
    name = u.Name
}
```
This matches how Slack itself renders mentions.

**2. Pipe-label format**

Slack sometimes encodes mentions as `<@U12345|display_name>` (with a pipe and label). The current regex `<@([UW][A-Z0-9]+)>` won't match these. A small tweak:
```go
regexp.MustCompile(`<@([UW][A-Z0-9]+)(?:\|[^>]+)?>`)
```

**3. Pre-compile regex**

`userMentionRegex` is compiled inside `filterSpecialChars` on every call. Moving it to a package-level `var` avoids repeated compilation:
```go
var userMentionRegex = regexp.MustCompile(`<@([UW][A-Z0-9]+)(?:\|[^>]+)?>`)
```
(Same applies to the URL and clean regexes, but that's a pre-existing issue — not your problem to fix here.)

**4. Minor: `slack.User` dependency in text package**

Threading `userMaps` into `filterSpecialChars` adds a `slack-go/slack` import to the text package, which was previously independent of Slack types. An alternative would be resolving mentions in `convertMessagesFromHistory` *before* calling `ProcessText`, keeping the text package stateless. But the placeholder approach is well-contained and consistent with how URLs are handled, so this is a trade-off I think is acceptable.

Overall this is in good shape. The test cases are thorough — particularly liked the "normal text same to user ID" and "invalid Slack user mention format" cases. Would be great to see this merged.

psyberck:
Thanks for the thorough comments. @chriscoey:). I will update the code accordingly.

psyberck:
Hey @chriscoey, I have refactored the implementations to resolve above concerns and feel like it's much cleaner. Let me know if there are more I can improve. Thanks!

korotovsky:
Thanks @psyberck — sorry for the delayed update.

I attempted to merge this and want to flag what I hit so you can decide on the rework path. Master has moved on quite a bit since 2026-03-10 and this PR no longer composes cleanly on top of it:

### 1. Conflict in \`pkg/handler/conversations.go\`

\`convertMessagesFromHistory\` and \`convertMessagesFromSearch\` were refactored (PR #269 / others) to use a \`userResolver\` instead of a local \`usersMap\` variable. Your two call sites:

\`\`\`go
Text: processText(msgText, usersMap.Users),
\`\`\`

need to become:

\`\`\`go
Text: processText(msgText, resolver.usersMap.Users),
\`\`\`

(or use \`ch.apiProvider.ProvideUsersMap().Users\` — same effect, slightly less current).

Also the same struct literal's \`Channel:\` field now uses \`fmt.Sprintf(\"%s (#%s)\", msg.Channel.ID, msg.Channel.Name)\` on master, not \`fmt.Sprintf(\"#%s\", msg.Channel.Name)\`. Take master's version.

### 2. \`TestUnitProcessText\` fails against current master

This is the bigger issue. The new test cases in this PR expect behavior that \`processText\` (as written) doesn't produce against the current \`text.ProcessText\`:

\`\`\`
processText(\"Hey <@B12348765> how are you?\")
    got:  \"Hey <@B12348765> how are you?\"
    want: \"Hey B12348765 how are you?\"

processText(\"I like <@  W87654321 > @U12345678 <U12345678> <@>\")
    got:  \"I like <@  W87654321 > @U12345678 <U12345678> <@>\"
    want: \"I like W87654321 U12345678 U12345678\"
\`\`\`

Root cause: \`text.UserMentionRegex\` is \`<@([UW][A-Z0-9]+)(?:\|[^>]+)?>\` — it only matches \`U\`/\`W\`-prefixed IDs in the strict \`<@…>\` form. So bot mentions (\`<@B…>\`), spaced-out mentions (\`<@ W… >\`), bare \`@USERID\`, and lone \`<USERID>\` all bypass the placeholder logic and pass through unchanged to \`text.ProcessText\`. And \`text.ProcessText\` itself has been narrowed since this PR was written (commits c298e5c, d3f7ca3, 2666c0f, etc.) — it now only does links/spaces/runes and no longer strips stray \`<@>\` brackets.

### What would unblock the merge

A version of \`processText\` that, after the user-mention substitution, also handles the cases the test expects. Roughly:

- bot mentions (\`<@B…>\`) → strip brackets, leave the ID
- malformed mentions (spaces, lone angle brackets, bare \`@USERID\`) → strip the noise, leave the ID
- unknown \`<@U…>\` → currently returns \`@U…\`, your tests assume this — fine to keep

Happy to re-review once that's in. The core idea — resolving mentions to display names in message text — is worth keeping.

psyberck:
@korotovsky I've updated the code, please take a look when you have time. Thanks!

Issue #206, opened 2026-02-18
Feature request: Add reactions_list tool


gJigsaw:
## Summary

Add a `reactions_list` tool that exposes the Slack API's `reactions.list` method.

## Use Case

When building daily summaries of "conversations I participated in," the current `conversations_search_messages` with `filter_users_with` only captures messages I sent/received. It misses conversations where I participated solely via emoji reactions.

The Slack API's `reactions.list` returns all items (messages, files, etc.) that the authenticated user has reacted to, which would fill this gap.

## Proposed Tool

```
Tool: reactions_list
Description: List items the authenticated user has reacted to
Parameters:
  - cursor: Pagination cursor
  - limit: Max items to return (default 100)
  - filter_date_after: Optional date filter (ISO-8601 or duration like -P1D)
```

## Reference

- Slack API docs: https://docs.slack.dev/reference/methods/reactions.list

Thanks for the great MCP server!

Issue #137, opened 2025-12-09
[Feature Request] Add support for message reactions (add/remove)


madamak:
## Feature Request

Add tools to add and remove emoji reactions to/from Slack messages.

## Use Case

When using AI agents to interact with Slack, it's common to want to:
- Acknowledge messages with reactions (👀, ✅, 👍, etc.)
- Mark messages as "seen" or "in progress" without posting a reply
- Remove reactions when status changes

This is particularly useful for:
- Triage workflows (react to acknowledge, then follow up)
- Status tracking without cluttering threads
- Quick acknowledgments that don't warrant a full message

### `reactions_add`
Add a reaction to a message.

**Parameters:**
- `channel` (required) - Channel ID
- `timestamp` (required) - Message timestamp
- `emoji` (required) - Emoji name without colons (e.g., `thumbsup`, `eyes`, `white_check_mark`)

### `reactions_remove`
Remove a reaction from a message.

**Parameters:**
- `channel` (required) - Channel ID  
- `timestamp` (required) - Message timestamp
- `emoji` (required) - Emoji name to remove

## Slack API Reference

- [`reactions.add`](https://api.slack.com/methods/reactions.add)
- [`reactions.remove`](https://api.slack.com/methods/reactions.remove)

Both methods require the `reactions:write` scope.

## Additional Consideration

A `reactions_get` tool could also be useful to retrieve existing reactions on a message, using [`reactions.get`](https://api.slack.com/methods/reactions.get) (requires `reactions:read` scope).

---

Thanks for this excellent MCP server! 🙏

bbelkacem:
+1

cbernard-dm:
+1

korotovsky:
I prioritized review of PR #141, @cbernard-dm @bbelkacem please consider to contribute there missing tool.

PR #239, opened 2026-03-14
feat: add reactions_get tool for detailed reaction data with user IDs

CJHwong:
Relates to #137 (completes the reactions tool family). See also #206 - `reactions_list` would require adding `reactions:read` scope; this tool covers the common case (who reacted to a specific message) without any new scope requirements.

## Problem

The existing reaction tools (`reactions_add`, `reactions_remove`) let agents manage reactions, but there's no way to read who reacted to a message. Issue #137 listed `reactions_get` as an additional consideration alongside the add/remove tools.

`conversations_history` includes reaction counts in message metadata, but not the per-user breakdown needed for use cases like emoji polls, voting, and tracking engagement.

## Solution

Add a read-only `reactions_get` tool that returns per-reaction user IDs for any message.

- Uses `conversations.replies` with `limit=1, inclusive=true` to fetch the target message
- Single API call works for top-level messages, thread parents, and thread replies alike
- Only requires `channels:history` scope (already required by other tools) - no new `reactions:read` scope needed
- Registered as a read-only tool (no env vars guard, enabled by default)
- Returns CSV with columns: `Emoji`, `Count`, `Users` (semicolon-separated user IDs)

**Parameters:**
- `channel_id` (required) - channel ID or name (`#general`, `@username`)
- `timestamp` (required) - message timestamp (`1234567890.123456`)

## Also fixes

README lines 91 and 101 referenced `SLACK_MCP_ADD_MESSAGE_TOOL` for `reactions_add`/`reactions_remove` - the actual env var is `SLACK_MCP_REACTION_TOOL`.

## Testing

Verified against real Slack workspace with xoxp token:

| Scenario | Result |
|----------|--------|
| Top-level message (no thread) | Correct reactions + user IDs |
| Thread parent (emoji poll, 8 reaction types, 11 voters) | Correct |
| Thread reply (2 reactions) | Correct |
| Thread reply (date vote, 6 reaction types, 10 voters each) | Correct |
| Message with no reactions | "No reactions on this message" |
| Invalid timestamp | "No message found at the specified timestamp" |

All existing tests pass (`TestValidToolNames`, `TestShouldAddTool_ReadOnly_EmptyEnabledTools`).


korotovsky:
technically there is a way to get them via message https://github.com/korotovsky/slack-mcp-server/blob/master/pkg/handler/conversations.go#L55 it has `Reactions` prop

CJHwong:
My initial plan was to add user IDs to the existing `Message.Reactions` field. But I found PR #91 where the format was changed from a full struct with user IDs to `emoji:count` to save LLM tokens in bulk history responses.

That made sense to me, so I went with a separate tool instead. It reads the raw `slack.Message` directly, so history stays lightweight and you only get the per-user breakdown when you ask for it.

PR #244, opened 2026-03-18
Load stale cache immediately on TTL expiry, refresh in background

### LINKED ISSUE #243: Support incremental cache refresh on TTL expiry instead of full rebuild
### Problem

When the disk cache expires (TTL), the server does a **full rebuild** — re-fetching all channels and users from the Slack API. On large Enterprise Grid workspaces (30K+ channels), this takes **~9 minutes** and blocks all tool calls with "cache is not ready yet" until complete.

This is related to #231 (startup rate-limiting) — both stem from the same full-rebuild approach.

### Current behavior

1. Cache file exists + within TTL → instant load from disk ✅
2. Cache expired → discard everything, call `GetChannels()` / `GetUsers()` for full list → ~9 min wait ❌
3. No middle ground

### Proposed behavior

When an expired cache exists on disk, **load it immediately** (mark `channelsReady = true`) so tool calls work instantly, then **refresh incrementally in the background**:

1. Load stale cache from disk on startup (tools become available immediately)
2. Kick off background goroutine to fetch delta/full refresh
3. Swap the in-memory snapshot atomically when refresh completes
4. Write updated cache to disk

The `client.userBoot` API already exists in the codebase (`pkg/provider/edge/client_boot.go`) and accepts `MinChannelUpdated` / `VersionTS` fields — these could potentially be used to fetch only channels updated since the cache was last written.

Even without a true delta API, just **loading stale cache first + refreshing in the background** would eliminate the cold-start penalty entirely. The worst case is a few minutes of slightly stale data, which is far better than 9 minutes of no data.

### Environment

- Workspace: Enterprise Grid (~30K channels)
- Cache TTL: 24h (`SLACK_MCP_CACHE_TTL=24h`)
- Cold rebuild time: ~9 minutes
- Version: v1.1.28

nhogade:
## Summary

When the disk cache exists but has expired (TTL), load it into the snapshot immediately so tools are available without waiting for a full API refetch. The fresh data is then fetched and the snapshot is atomically swapped when complete.

On large Enterprise Grid workspaces (30K+ channels), the full cache rebuild takes **~9 minutes**. Previously, all tool calls blocked with "cache is not ready yet" during this period. Now they work instantly with slightly stale data while the refresh happens transparently.

## Changes

Single file change (`pkg/provider/api.go`) — both `refreshUsersInternal` and `refreshChannelsInternal`:

- **Before**: expired cache → discard → block on full API fetch → set `ready = true`
- **After**: expired cache → load stale data → set `ready = true` → fetch fresh data → atomically swap snapshot

The key insight is that `ProvideUsersMap()` and `ProvideChannelsMaps()` use `atomic.Pointer` loads (no lock needed), so tool calls can read the stale snapshot while the refresh goroutine holds the write lock for the API fetch.

## How it works

1. Cache file exists but TTL expired
2. Stale data loaded into `atomic.Pointer` snapshot → `channelsReady = true` / `usersReady = true`
3. `IsReady()` returns `true` → server starts accepting tool calls immediately
4. Function continues to fetch fresh data from Slack API (still under write lock — prevents concurrent refreshes)
5. When fresh data arrives, `atomic.Pointer.Store()` swaps the snapshot seamlessly
6. Updated cache written to disk

## Testing

- All existing tests pass (`go test ./pkg/provider/...`)
- Verified `go build ./...` succeeds

Fixes #243

### DIFF
--- a/pkg/provider/api.go
+++ b/pkg/provider/api.go
@@ -782,37 +782,45 @@ func (ap *ApiProvider) refreshUsersInternal(ctx context.Context, force bool) err
 					zap.String("cache_file", ap.usersCachePath))
 			} else {
 				// Check cache TTL using file modification time
-				cacheValid := true
+				cacheExpired := false
 				if ap.cacheTTL > 0 {
 					if fileInfo, err := os.Stat(ap.usersCachePath); err == nil {
 						cacheAge := time.Since(fileInfo.ModTime())
 						if cacheAge > ap.cacheTTL {
-							ap.logger.Info("Users cache expired, will refetch",
+							cacheExpired = true
+							ap.logger.Info("Users cache expired, loading stale data for immediate availability",
 								zap.Duration("cache_age", cacheAge),
 								zap.Duration("ttl", ap.cacheTTL),
 								zap.String("cache_file", ap.usersCachePath))
-							cacheValid = false
 						}
 					}
 				}
 
-				if cacheValid {
-					// Build new snapshot from cache
-					newSnapshot := &UsersCache{
-						Users:    make(map[string]slack.User, len(cachedUsers)),
-						UsersInv: make(map[string]string, len(cachedUsers)),
-					}
-					for _, u := range cachedUsers {
-						newSnapshot.Users[u.ID] = u
-						newSnapshot.UsersInv[u.Name] = u.ID
-					}
-					ap.usersSnapshot.Store(newSnapshot)
+				// Always load existing cache into snapshot for immediate availability.
+				// This ensures tools are usable instantly even when the cache is stale.
+				newSnapshot := &UsersCache{
+					Users:    make(map[string]slack.User, len(cachedUsers)),
+					UsersInv: make(map[string]string, len(cachedUsers)),
+				}
+				for _, u := range cachedUsers {
+					newSnapshot.Users[u.ID] = u
+					newSnapshot.UsersInv[u.Name] = u.ID
+				}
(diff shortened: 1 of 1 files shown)


chriscoey (commented):
Good approach to the cold-start problem. The key change — always loading the existing cache into the snapshot and setting `usersReady`/`channelsReady` *before* falling through to the API fetch — means tools are available instantly with slightly stale data. The atomic snapshot swap on refresh completion keeps concurrent access safe.

The pattern is applied symmetrically across both users and channels caches, which makes the change easy to follow. An additional benefit: if the API fetch after loading stale data fails (e.g., rate limited), the stale data is still available — strictly better than the current behavior where an expired cache is discarded before re-fetching.

This would be a big improvement for large workspaces where cold refreshes take minutes.

gkatz2:
I want to second @chriscoey's comment above. I run this server on a large Enterprise Grid workspace with a long cache TTL and a persistent volume mount so the cache survives container restarts. This setup gives me sub-second startups throughout the day, but the first start after the TTL expires still blocks for 5–10 minutes while the full API rebuild runs.

This PR would eliminate that last cold-start penalty. I would love to see this get merged. Happy to help test if that would be useful.

korotovsky:
Closing in favor of #225, which targets the same stale-while-revalidate cache strategy with a more complete implementation (`atomic.Bool` ready flags that fix a real data race, refresh coalescing via CompareAndSwap, removal of the stdio polling loop). #225 needs a rebase but is the substantively complete version. Thanks for the contribution!

PR #249, opened 2026-03-24
fix: don't block stdio transport on cache warmup

mark-liu:
## Problem

The stdio transport blocks serving until `IsReady()` returns true, which requires both the users and channels caches to be fully populated. On large workspaces (Kubernetes Slack has 225k+ members), the user cache load takes several seconds — long enough that MCP clients with startup timeouts (like Claude Code) fail the connection before the server is ready.

The SSE and HTTP transports don't have this problem — they serve immediately and log "still warming up caches" while the background goroutine populates.

## Fix

Remove the blocking `IsReady()` loop for stdio and match the SSE/HTTP behaviour: serve immediately, let caches warm in the background.

Tools that depend on the user cache (DM channel naming, user search, display name resolution in message history) gracefully degrade by showing user IDs until the cache is ready.

## Testing

Tested against Kubernetes Slack (225,807 members, 555 channels):
- **Before**: stdio transport times out during cache load (~4s blocked)
- **After**: stdio transport serves immediately, caches warm in background

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 /npm/slack-mcp-server/LICENSE
 /npm/slack-mcp-server/README.md
 docker-compose.release.yml
+bin/
--- a/cmd/slack-mcp-server/main.go
+++ b/cmd/slack-mcp-server/main.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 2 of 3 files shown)


mark-liu:
Superseded by #259 (squashed and rebased with #256).

Issue #255, opened 2026-03-26
Duplicate AuthTest API call on every startup


mark-liu:
Each startup path (`newWithXOXP`, `newWithXOXC`) calls `AuthTest()` twice:

1. `validateAuthAndGetTeamID()` — calls AuthTest to get the TeamID for cache paths
2. `NewMCPSlackClient()` — calls AuthTest again to configure the edge client

The first response is discarded after extracting TeamID. The second rebuilds the same data.

When running multiple instances (e.g. separate workspaces), this doubles the auth API calls at startup for no reason, increasing the chance of hitting Slack rate limits.

Fix: return the full `AuthTestResponse` from `validateAuthAndGetTeamID` and pass it into `NewMCPSlackClient` so it can skip the redundant call.

I have a fix ready — will open a PR together with #254.

PR #256, opened 2026-03-26
Atomic cache writes and deduplicate AuthTest on startup

### LINKED ISSUE #254: Cache writes are not atomic — concurrent instances can corrupt JSON files
When multiple MCP server instances share the same cache directory (e.g. Claude Code launching several slack-mcp-server processes simultaneously), the `os.WriteFile` calls in `refreshUsersInternal` and `refreshChannelsInternal` can leave partial JSON visible to concurrent readers.

The cache files are 300-460MB on large workspaces. A concurrent read during a write returns truncated JSON, which causes cache load failures on next startup.

* `pkg/provider/api.go` line 871: `os.WriteFile(ap.usersCachePath, data, 0644)`
* `pkg/provider/api.go` line 991: `os.WriteFile(ap.channelsCachePath, data, 0644)`

Fix: write to a temp file in the same directory, then `os.Rename` to the target path. Rename is atomic on POSIX filesystems.

I have a fix ready — will open a PR.

mark-liu:
Fixes #254 and #255.

Two fixes for running multiple MCP server instances concurrently (e.g. Claude Code launching separate instances per workspace):

* Cache files are now written atomically via write-to-temp-then-rename. Previously `os.WriteFile` on 300-460MB JSON files could leave partial data visible to concurrent readers.

* `validateAuthAndGetTeamID` now returns the full `AuthTestResponse` which is passed into `NewMCPSlackClient`, eliminating a redundant `AuthTest` API call per startup (2 per instance down to 1). Reduces rate limit pressure when multiple instances start simultaneously.

No new dependencies. Backwards compatible — `NewMCPSlackClient` accepts `nil` for `cachedAuth` and falls back to calling `AuthTest` itself.

### DIFF
--- a/pkg/provider/api.go
+++ b/pkg/provider/api.go
@@ -114,12 +114,12 @@ func getMinRefreshInterval() time.Duration {
 // This ensures tokens are valid before proceeding and enables cache namespacing
 // to prevent cache contamination when using multiple Slack workspaces.
 // Returns an error if authentication fails - the server should not start with invalid credentials.
-func validateAuthAndGetTeamID(authProvider auth.Provider, logger *zap.Logger) (string, error) {
+func validateAuthAndGetTeamID(authProvider auth.Provider, logger *zap.Logger) (string, *slack.AuthTestResponse, error) {
 	xoxpToken := os.Getenv("SLACK_MCP_XOXP_TOKEN")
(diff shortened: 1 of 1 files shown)


mark-liu:
Superseded by #259 (squashed and rebased with #249).

PR #259, opened 2026-03-26
Non-blocking stdio startup, atomic cache writes, deduplicate AuthTest

mark-liu:
Three related fixes for reliability when running multiple concurrent
MCP server instances (e.g. one per Slack workspace):

* stdio transport no longer blocks on cache warmup — serves
  immediately while caches build in the background, matching the
  existing SSE/HTTP behaviour. Large workspaces (100K+ users)
  previously timed out before the server started listening.

* Cache files written atomically via temp-file-and-rename,
  preventing concurrent readers from seeing partial/corrupt JSON.

* AuthTest response from initial validation passed through to
  NewMCPSlackClient, eliminating a redundant API call on every
  startup. Small random jitter (0-3s) staggers concurrent
  instance auth requests.

Supersedes #249 and #256 (squashed and rebased).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 /npm/slack-mcp-server/LICENSE
 /npm/slack-mcp-server/README.md
 docker-compose.release.yml
+bin/
--- a/cmd/slack-mcp-server/main.go
+++ b/cmd/slack-mcp-server/main.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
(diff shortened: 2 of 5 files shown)


korotovsky:
Closing in favor of #277, which landed as the canonical fix for the non-blocking stdio startup (issue #271). #277 is smaller and more focused; this PR mixes three concerns (stdio + atomic cache writes + AuthTest dedup + jitter). The atomic-cache-write and AuthTest-dedup parts are valuable — please consider splitting them into separate focused PRs and we'll happily review. Thanks for the work!

PR #133, opened 2025-11-25
Add get_permalink tool to generate Slack message URLs

alexgarcia-wq:
## Summary
Implements a new MCP tool `get_permalink` that constructs Slack permalinks from channel IDs and message timestamps. This addresses feature request #100.

## Motivation
As requested in #100 by @cmarken, this enables users to include permalink sources when summarizing Slack conversations. This is particularly valuable for daily standup summaries, citations, and quick navigation back to important messages.

## Changes
- **Added** `FormatPermalinkTimestamp()` helper function in `pkg/text/text_processor.go` to convert Slack timestamp format (`1234567890.123456`) to permalink format (`1234567890123456`)
- **Added** `GetPermalinkHandler()` method in `pkg/handler/conversations.go` that:
  - Accepts `channel_id` and `message_ts` parameters
  - Retrieves workspace name via `Slack().AuthTest()`
  - Constructs and returns the permalink URL
- **Added** `parseParamsToolGetPermalink()` validation method
- **Registered** the new tool in `pkg/server/server.go` with proper parameter descriptions

## Usage Example
```go
get_permalink({
  channel_id: "C1234567890",  // or D1234567890 for DMs
  message_ts: "1234567890.123456"
})
```

Returns:
```
https://{workspace}.slack.com/archives/C1234567890/p1234567890123456
```

## Testing
Tested locally with real Slack workspace messages. The tool successfully:
- Extracts workspace name from auth response
- Formats timestamps correctly
- Generates valid, clickable Slack permalinks

## Related Issues
Closes #100

cc @cmarken @Brandfirst @freddyhaddad

Issue #100, opened 2025-09-23
Support to get the permalink to a message


cmarken:
As a daily summarization agent I’d like to get access to the permalink of messages so that I can include it as the source of my summarization. 

This would enable me as a user to easily read up on the details when I find an area I want to dive deeper into and understanding.

Brandfirst:
I support this

freddyhaddad:
I support this

alexgarcia-wq:
I've implemented this feature! 🎉

Added a new `get_permalink` tool that generates Slack message permalinks from channel IDs and message timestamps.

**Usage:**
```
get_permalink({
  channel_id: "C1234567890",  // or D1234567890 for DMs
  message_ts: "1234567890.123456"
})
```

**Returns:**
```
https://{workspace}.slack.com/archives/{channel_id}/p{timestamp}
```

This enables exactly what @cmarken requested - the ability to include permalink sources in summarization workflows.

I'll be submitting a PR shortly. @Brandfirst @freddyhaddad - this should address your needs as well!

korotovsky:
Could you guys elaborate a little more what this requires a tool and why this can not be handled with simple prompt to LLM by that agent? From the implementation perspective the tool does a + b and a little of formatting...

korotovsky:
Resolved — `Permalink` is now included in message-returning tools (`conversations_history`, `conversations_replies`, `conversations_search_messages`) via #265 and follow-up work. Closing as completed.

PR #265, opened 2026-03-30
Include channel ID and permalink in search results

tmastny:
Addresses #100.

Search results from conversations_search_messages only includes the channel display name, like #general. The Slack search API returns both Channel.ID and Permalink on every SearchMessage, but they are not included in convertMessagesFromSearch.

I often want permalinks to the message to add to docs or to open the thread myself, but I can't reliably get the LLM agent to reconstruct the link. This makes it possible.

Changes:
- Channel column now includes the ID: `C0515UGHR0R (#general)`
- New `Permalink` column passes through `msg.Permalink` from the API response

This is similar to #133, but I think worthwhile on its own. Instead of adding a new tool to construct permalinks (which is still useful!) this passes permalinks through when available.

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -52,6 +52,7 @@ type Message struct {
 	ThreadTs      string `json:"ThreadTs"`
 	Text          string `json:"text"`
 	Time          string `json:"time"`
+	Permalink     string `json:"permalink,omitempty"`
 	Reactions     string `json:"reactions,omitempty"`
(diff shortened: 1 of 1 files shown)


korotovsky:
Merging — small, useful change. `convertMessagesFromSearch` now includes channel ID alongside name and adds `Permalink` (omitempty). Tiny surface, helps downstream consumers correlate search hits with conversations. Thanks!

PR #269, opened 2026-04-03
Add targeted user cache patch on miss

### LINKED ISSUE #268: User cache miss returns raw IDs instead of resolving via users.info
When a user ID isn't in the in-memory cache (e.g., a new hire or Slack Connect guest who joined after the cache was built), there's no recovery path. Message rendering degrades to showing raw user IDs (`U01234567`) instead of display names, and `paramFormatUser` returns a hard error that fails the entire tool call.

The channel cache has a miss-recovery pattern (`ForceRefreshChannels` + retry), but the user cache has no equivalent. On Enterprise Grid workspaces with ~50K users and long cache TTLs, the user data can be hours stale with no way to resolve individual misses at runtime.

**Proposed fix:**

Add a `PatchUser` method that calls `users.info` for a single user ID (O(1) API call), then patches the in-memory snapshot atomically. Wire it into:
- `paramFormatUser` for search filter resolution (ID-based misses)
- `convertMessagesFromHistory` and `convertMessagesFromSearch` for message rendering

This avoids the cost of a full cache rebuild (~50K users) for a single missing entry. Disk persistence is skipped since the next full refresh cycle will persist it.

Channel name-to-ID targeted lookup is not viable (Slack has no "get channel by name" API), so this is scoped to users only.

gkatz2:
## Summary
- When a user ID isn't in the in-memory cache, fetch it via `users.info` (O(1) API call) and patch the snapshot atomically, instead of showing raw IDs or failing the tool call
- Wire recovery into `paramFormatUser` (search filters), `convertMessagesFromHistory`, and `convertMessagesFromSearch` (message rendering)
- Introduce a `userResolver` that encapsulates cache lookup, on-demand fetch, and failed-lookup dedup for message batches
- On Enterprise Grid (~50K users), this avoids a full cache rebuild (minutes of API calls) for a single missing entry

Fixes #268

## Test plan
- [x] Unit tests for `PatchUser` (success, API error, empty/nil result, snapshot immutability, overwrite)
- [x] `go build ./...` passes
- [x] `make test` passes (all existing + new tests)
- [ ] Manual test: run MCP server, query messages from a user not in cache, verify name resolves
- [ ] Upstream CI

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -527,7 +527,7 @@ func (ch *ConversationsHandler) ConversationsHistoryHandler(ctx context.Context,
 
 	ch.logger.Debug("Fetched conversation history", zap.Int("message_count", len(history.Messages)))
 
-	messages := ch.convertMessagesFromHistory(history.Messages, params.channel, params.activity)
+	messages := ch.convertMessagesFromHistory(ctx, history.Messages, params.channel, params.activity)
 
 	if len(messages) > 0 && history.HasMore {
 		messages[len(messages)-1].Cursor = history.ResponseMetaData.NextCursor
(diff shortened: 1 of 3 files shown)


korotovsky:
Thanks @gkatz2 — clean fix and good test coverage. Closes #268. Merging.

(The most recent integration-tests run failed but it looks unrelated to this change; will re-run on `master` after merge.)

PR #190, opened 2026-02-07
feat: extract text content from message blocks in conversations_history

derodero24:
## Summary
Extract text from Slack Block Kit structures in messages retrieved via `conversations_history` and `conversations_replies`, so that block-only messages (emails, bot notifications, app messages) are no longer returned empty.

Closes #186

## Problem
Messages using Block Kit format (e.g., Slack Email integration, Grafana/Datadog alerts) have an empty `Text` field. The existing `AttachmentsTo2CSV` extracts text from attachment fields but ignores:
- Top-level `blocks` array
- `attachments[].blocks` array

## Solution
1. **`BlocksToText`** function in `pkg/text/text_processor.go` extracts text from common block types:
   - `header.text`
   - `section.text` and `section.fields`
   - `rich_text` (sections, lists, quotes, preformatted)
   - `context.elements[].text`
2. **Top-level blocks**: When `msg.Text` is empty, fall back to `BlocksToText(msg.Blocks)` — avoids duplication since Slack typically populates `text` as a plaintext fallback of blocks
3. **Attachment blocks**: `AttachmentToText` now also extracts `att.Blocks`

## Changes
- `pkg/text/text_processor.go`: Add `BlocksToText` and rich text helper functions; call `BlocksToText` in `AttachmentToText`
- `pkg/handler/conversations.go`: Use block text as fallback in `convertMessagesFromHistory` and `convertMessagesFromSearch`
- `pkg/text/text_processor_test.go`: Add 15 test cases covering all supported block types, edge cases, and attachment integration

## Testing
- All existing unit tests pass (`go test -run TestUnit ./...`)
- New tests cover: empty blocks, header, section (with text/fields/nil text), context, rich_text (text/link/broadcast), multiple blocks combined, non-text blocks skipped, attachment with blocks only, attachment with both text and blocks

PR #275, opened 2026-04-07
feat: render Slack Block Kit content in message text

alangmartini:
## Summary

Bot messages (workflow automations, monitoring bots, etc.) often store their content exclusively in the `blocks` array rather than the `text` field. Previously this content was silently dropped, causing these messages to appear empty or with only a summary line in the CSV output.

This PR adds block content rendering so the full message data is available to MCP clients.

### Changes

- **New file `pkg/text/blocks.go`**: `BlocksToText()` function that traverses Block Kit structures and converts them to plain text. Supports:
  - `section` blocks (text + fields)
  - `header` blocks
  - `rich_text` blocks (sections, lists, quotes, preformatted)
  - `context` blocks (text + image alt text)
  - Rich text section elements: text, links, emoji, channels, users, broadcasts, dates, colors, user groups

- **`pkg/handler/conversations.go`**: Both `convertMessagesFromHistory` and `convertMessagesFromSearch` now call `MergeBlocksWithText()` to include block content in the text output. `MergeBlocksWithText` avoids duplication when both `text` and `blocks` contain the same content.

### Affected tools

- `conversations_history`
- `conversations_replies`
- `conversations_unreads`
- `conversations_search_messages`

### Example

Before (bot message with block-only content):
```
Text: "Details page 1 10 rows"
```

After:
```
Text: "Details page 1 10 rows\ncluster_id: abc123\nalert_types: Low Memory\nauto_upgrade: yes\n..."
```

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -1489,7 +1489,7 @@ func (ch *ConversationsHandler) convertMessagesFromHistory(slackMessages []slack
 			continue
 		}
 
-		msgText := msg.Text + text.AttachmentsTo2CSV(msg.Text, msg.Attachments)
+		msgText := text.MergeBlocksWithText(msg.Text, msg.Blocks) + text.AttachmentsTo2CSV(msg.Text, msg.Attachments)
 
 		var reactionParts []string
 		for _, r := range msg.Reactions {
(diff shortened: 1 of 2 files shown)


korotovsky:
Closing in favor of #190, which targets the same goal (extracting Block Kit content into the rendered message text) with broader coverage: top-level blocks, attachment blocks, and rich text elements, plus 15 test cases. Thanks for the work — please review #190 if you'd like to add suggestions there.

PR #277, opened 2026-04-10
fix: non-blocking stdio startup and --no-cache flag

### LINKED ISSUE #271: Server blocks MCP initialize response while caching users/channels on startup
## Summary

On large Slack workspaces, the server caches all users and channels **before** responding to the MCP `initialize` request. This causes MCP clients with connection timeouts (like Claude Code's 60-second default) to time out and drop the server.

## Reproduction

On a workspace with ~12.8k users and ~43k channels:

1. Start the binary via stdio
2. Send MCP `initialize` request immediately
3. Observe: the server authenticates to Slack, then starts caching users (~11s) and channels (3-5 min) BEFORE responding to `initialize`

Stderr shows the progression:
```
{"level":"info","message":"Using Bot token authentication","token_type":"xoxb"}
{"level":"info","message":"Authenticated to Slack","team":"...","user":"..."}
{"level":"info","message":"Caching users collection..."}
{"level":"info","message":"Wrote users to cache","count":12882}  // ~11s
{"level":"info","message":"Caching channels collection..."}
// ... 3-5 minutes of silence, no MCP response on stdout ...
```

## Expected Behavior

The server should respond to `initialize` (and `tools/list`) immediately, then cache users/channels in the background. The cache should only block tool calls that need the data (like `channels_list`), not the MCP protocol handshake.

## Workaround

Pre-warm the cache by running the binary once before the MCP client connects. The cache files at `~/.cache/slack-mcp-server/` are reused on subsequent runs.

## Environment

- slack-mcp-server version: 1.2.3
- Binary: `slack-mcp-server-linux-amd64`
- Slack workspace: ~12.8k users, ~43k channels
- MCP client: Claude Code 2.1.91 (via GitHub Actions)

mustafa89:
## Summary

Fixes #271 — the stdio transport blocks the MCP `initialize` response until users/channels caches are fully loaded. On large workspaces (12k+ users, 43k+ channels) this takes 3-5 minutes, causing MCP clients with connection timeouts to drop the server.

### 1. Non-blocking stdio startup (default behavior change)

Removed the busy-wait loop in `main.go` that blocked stdio from starting until `IsReady()` returned true:

```go
// BEFORE — blocks until caches are loaded
case "stdio":
    for {
        if ready, _ := p.IsReady(); ready {
            break
        }
        time.Sleep(100 * time.Millisecond)
    }
    s.ServeStdio()
```

```go
// AFTER — starts immediately, caches load in background
case "stdio":
    if ready, _ := p.IsReady(); !ready && !noCache {
        logger.Info("Slack MCP Server is still warming up caches, starting server anyway")
    }
    s.ServeStdio()
```

This matches the existing SSE/HTTP behavior, which already starts serving before caches are ready. Tool calls made before caches load return a graceful "not ready" error via the existing `buildErrorRecoveryMiddleware`, allowing the LLM to retry.

### 2. `--no-cache` CLI flag (opt-in)

For environments that only use channel/user IDs (never `#channel-name` or `@username` lookups), the `--no-cache` flag skips cache loading entirely:

```
slack-mcp-server --no-cache --transport stdio
```

This marks both `usersReady` and `channelsReady` as true immediately via a new `SkipCache()` method on `ApiProvider`, without loading any data. Startup is instant regardless of workspace size.

### Files changed

| File | Change |
|------|--------|
| `cmd/slack-mcp-server/main.go` | Add `--no-cache` flag, remove stdio busy-wait loop, add warm-up log message |
| `pkg/provider/api.go` | Add `SkipCache()` method |

## Testing

Tested on a workspace with 823 users and 1,418 channels:

| Mode | Startup time | Notes |
|------|-------------|-------|
| `--no-cache` | <1s | Instant, no cache loading |
| Default (no flag) | <1s | MCP handshake immediate, caches load in background |
| Original (upstream) | 30s+ timeout | Blocked on channel cache refresh |

### DIFF
--- a/cmd/slack-mcp-server/main.go
+++ b/cmd/slack-mcp-server/main.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/korotovsky/slack-mcp-server/pkg/provider"
 	"github.com/korotovsky/slack-mcp-server/pkg/server"
@@ -23,10 +22,12 @@ var defaultSsePort = 13080
 func main() {
 	var transport string
 	var enabledToolsFlag string
+	var noCache bool
 	flag.StringVar(&transport, "t", "stdio", "Transport type (stdio, sse or http)")
 	flag.StringVar(&transport, "transport", "stdio", "Transport type (stdio, sse or http)")
 	flag.StringVar(&enabledToolsFlag, "e", "", "Comma-separated list of enabled tools (empty = all tools)")
 	flag.StringVar(&enabledToolsFlag, "enabled-tools", "", "Comma-separated list of enabled tools (empty = all tools)")
+	flag.BoolVar(&noCache, "no-cache", false, "Skip user/channel cache loading on startup for faster initialization. Lookups by #channel-name or @username will not work; use channel/user IDs instead.")
 	flag.Parse()
 
 	if enabledToolsFlag == "" {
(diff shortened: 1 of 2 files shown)


korotovsky:
Merging — canonical fix for #271. Smaller, focused approach (removes the busy-wait loop, adds `SkipCache()` provider method, and a `--no-cache` flag for ID-only deployments). Idiomatic Go, mirrors existing SSE/HTTP behavior. Thanks!

PR #279, opened 2026-04-15
Include file names in AttachmentIDs for message responses

### LINKED ISSUE #260: [Feature Request] Include file_id in slack_read_thread response for attachments
## Summary

`slack_read_thread` returns only **file names** for attachments, but does not include the **file_id (Fxxxxxxxxxx)** needed to call `attachment_get_data`. This makes it impossible to programmatically read file contents from thread messages.

## Current Behavior

`slack_read_thread` (response_format="detailed") output:

```
Files: contract_review.docx
```

- Only the file title/name is shown
- No file_id (e.g., `F08XXXXXXXX`)

## Expected Behavior

```
Files: contract_review.docx (file_id: F08ABC1234)
```

Or as structured fields:

```
Files:
  - name: contract_review.docx
    file_id: F08ABC1234
    filetype: docx
    size: 123456
```

## Affected Tools

- `slack_read_thread` (primary impact)
- `slack_read_channel` (likely same issue)
- `slack_search_public_and_private` (likely same issue)

## Use Case

We have a skill that automatically reviews legal documents (contracts) uploaded as attachments in Slack threads. Without file_id, we cannot call `attachment_get_data` to download and analyze the files. Currently, users must manually paste the document text as a workaround.

## Root Cause

The Slack API's `conversations.replies` response includes a `files` array in each message with the `id` field. The MCP server's response formatting strips this information and only passes through the file title/name. Exposing the `id` field in the formatted output would resolve this.

## Workaround

Currently asking users to either:
1. Paste document text directly into the thread
2. Share the file's permalink (which sometimes contains the file_id in the URL)

bkbaheti:
## Summary

Fixes #260

- Include file names alongside file IDs in the `AttachmentIDs` field of message responses from `slack_read_thread` and `slack_read_channel`
- Previously: `F08ABC1234,F08XYZ5678` — raw IDs with no way to identify which file is which
- Now: `F08ABC1234 (contract.pdf), F08XYZ5678 (report.xlsx)` — IDs with filenames so consumers can selectively download relevant attachments via `attachment_get_data`
- Updated `attachment_get_data` tool description to reference the new format

## Notes

- `convertMessagesFromSearch` is not updated because the upstream `slack-go` `SearchMessage` struct does not include a `Files` field — the Slack search API may return file data but it's silently dropped during unmarshalling. This would need an upstream fix in `slack-go/slack`.
- Falls back to ID-only format if a file has no name set.

## Test plan

- [ ] Read a channel/thread containing messages with file attachments
- [ ] Verify `AttachmentIDs` field now shows `F... (filename)` format
- [ ] Verify `attachment_get_data` still works when called with the file ID extracted from the new format
- [ ] Verify messages without attachments are unaffected

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -1507,9 +1507,13 @@ func (ch *ConversationsHandler) convertMessagesFromHistory(slackMessages []slack
 
 		var attachmentIDs []string
 		for _, f := range msg.Files {
-			attachmentIDs = append(attachmentIDs, f.ID)
+			if f.Name != "" {
+				attachmentIDs = append(attachmentIDs, fmt.Sprintf("%s (%s)", f.ID, f.Name))
+			} else {
+				attachmentIDs = append(attachmentIDs, f.ID)
+			}
 		}
-		attachmentIDsStr := strings.Join(attachmentIDs, ",")
+		attachmentIDsStr := strings.Join(attachmentIDs, ", ")
 
 		messages = append(messages, Message{
 			MsgID:         msg.Timestamp,
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -231,7 +231,7 @@ func NewMCPServer(provider *provider.ApiProvider, logger *zap.Logger, enabledToo
 			mcp.WithReadOnlyHintAnnotation(true),
 			mcp.WithString("file_id",
 				mcp.Required(),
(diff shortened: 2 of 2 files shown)


korotovsky:
Merging — closes #260. Tiny output enrichment that adds the file name alongside the file ID in attachments, making downstream `attachment_get_data` calls easier to correlate. Thanks!

Issue #278, opened 2026-04-14
Parentheses replacement in AttachmentToText breaks markdown link format


tim-watcha:
## Problem

`AttachmentToText()` unconditionally replaces all `(` with `[` and `)` with `]`:

https://github.com/korotovsky/slack-mcp-server/blob/d380003/pkg/text/text_processor.go#L57-L58

```go
result = strings.ReplaceAll(result, "(", "[")
result = strings.ReplaceAll(result, ")", "]")
```

This prevents using standard markdown link syntax `[text](url)` in the output — it gets transformed into `[text][url]`, which is not a valid link format for downstream consumers (LLMs, markdown renderers, etc.).

For example, when an attachment has a `TitleLink`, the natural output would be:

```
Title: [Read timeout on endpoint URL](https://sentry.io/issues/123)
```

But instead it becomes:

```
Title: [Read timeout on endpoint URL][https://sentry.io/issues/123]
```

## Suggestion

Consider removing the parentheses replacement, or at minimum, exempting URL-containing patterns from the replacement. The original intent seems to be CSV-safety, but the current MCP output format (`;`-delimited parts) does not appear to require this escaping.

korotovsky:
Resolved by #272 (extracts Fields and TitleLink from attachments) and #281 (removes the legacy whitelist filter that was stripping characters), both now merged. The `(` → `[` substitution that was breaking markdown links has been removed.

PR #283, opened 2026-04-21
fix(text): remove parentheses replacement in AttachmentToText

Will-hxw:
## Summary

Removes the unconditional replacement of `(` with `[` and `)` with `]` in `AttachmentToText()`. This was breaking markdown link format `[text](url)` by converting it to `[text][url]`.

## Problem

The `AttachmentToText` function was replacing all parentheses with square brackets:

```go
result = strings.ReplaceAll(result, "(", "[")
result = strings.ReplaceAll(result, ")", "]")
```

This caused markdown links like `[Read timeout](https://sentry.io/issues/123)` to become `[Read timeout][https://sentry.io/issues/123]`, which is not valid markdown.

## Fix

Removed both replacement lines. The original CSV-safety escaping is unnecessary since the output format uses `;` as delimiter, not comma.

## Test plan

- [x] Go build passes on cmd package
- [x] Manual verification that markdown links remain intact in output

### DIFF
--- a/pkg/text/text_processor.go
+++ b/pkg/text/text_processor.go
@@ -44,8 +44,6 @@ func AttachmentToText(att slack.Attachment) string {
 	result = strings.ReplaceAll(result, "\n", " ")
 	result = strings.ReplaceAll(result, "\r", " ")
 	result = strings.ReplaceAll(result, "\t", " ")
-	result = strings.ReplaceAll(result, "(", "[")
-	result = strings.ReplaceAll(result, ")", "]")
 	result = strings.TrimSpace(result)
 
 	return result
--- a/pkg/text/text_processor.go
+++ b/pkg/text/text_processor.go
@@ -19,7 +19,7 @@ func AttachmentToText(att slack.Attachment) string {
 
 	if att.Title != "" {
 		if att.TitleLink != "" {
-			parts = append(parts, fmt.Sprintf("Title: %s (%s)", att.Title, att.TitleLink))
+			parts = append(parts, fmt.Sprintf("Title: [%s](%s)", att.Title, att.TitleLink))
 		} else {
 			parts = append(parts, fmt.Sprintf("Title: %s", att.Title))
 		}
--- a/pkg/text/text_processor.go
+++ b/pkg/text/text_processor.go
@@ -18,7 +18,11 @@ func AttachmentToText(att slack.Attachment) string {
 	var parts []string
 
 	if att.Title != "" {
-		parts = append(parts, fmt.Sprintf("Title: %s", att.Title))
+		if att.TitleLink != "" {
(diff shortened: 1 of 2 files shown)


Will-hxw:
Hi @140am @korotovsky @JAORMX , sorry to bother you. This PR has been open for a while, so I wanted to kindly check whether you might have time to review it or let me know if any changes are needed.I’m happy to make adjustments if required. Thanks!

140am:
tested this a bit more against a real slack attachment payload and i think this is directionally right but still incomplete

the simple unit-level case `[Read timeout](https://sentry.io/issues/123)` is fixed by removing the paren replacement, but live slack returned the attachment title as:

`[Read timeout](<https://sentry.io/issues/123>)`

with the url wrapped in `<...>`

with the current PR, that still comes back through the MCP as:

`Title: Read timeouthttps://sentry.io/issues/123`

so i think the markdown parser also needs to handle the slack-shaped `[text](<url>)` form, something like:

```go
regexp.MustCompile(`\[([^\]]+)\]\(<?(https?://[^)>]+)>?\)`)
```

i also tested the separate `title` + `title_link` attachment case, since #278 mentions `TitleLink`. that still comes back as just `Title: Read timeout`, because `AttachmentToText` does not currently read `slack.Attachment.TitleLink`

so the fix likely needs two pieces:

1. keep this paren replacement removal
2. support `[text](<url>)` in `filterSpecialChars`
3. optionally render `TitleLink` as `[Title](TitleLink)` before `ProcessText`

would also be good to add regression tests for both live-shaped cases

Will-hxw:
Good catch on the \`<...>\` wrapper format! I've pushed an update that handles the angle-bracket markdown link style \`[text](<url>)\`.

Also added \`TitleLink\` support in \`AttachmentToText\` so \`Title: text (url)\` format is included.

Both changes are on \`fix/278-parentheses-markdown-link\` branch with test coverage. Let me know if there are other edge cases to cover.

140am:
the `[text](<url>)` live slack case looks fixed now

i reran the same live attachment fixture through the updated PR and the markdown-title case now comes back cleanly:

`Title: https://sentry.io/issues/123 - Read timeout, Text: markdown attachment fixture`

one remaining issue: the separate `title` + `title_link` case still comes back with punctuation leakage:

`Title: Read timeout https://sentry.io/issues/123); Text: title_link attachment fixture`

looks like this is from formatting it as `Title: %s (%s)`. changing that to markdown form should let the existing parser normalize it consistently:

```go
parts = append(parts, fmt.Sprintf("Title: [%s](%s)", att.Title, att.TitleLink))
```

i tested that locally and against the live fixture, and it produced:

`Title: https://sentry.io/issues/123 - Read timeout, Text: title_link attachment fixture`

would also add a regression test for `slack.Attachment{Title, TitleLink}` since the current new tests cover `[text](<url>)` but not the `TitleLink` path

PR #284, opened 2026-04-26
fix: return plain confirmation from conversations_add_message

clwluvw:
The handler was re-fetching the just-posted message via conversations.history and returning it as CSV. Slack's history endpoint has propagation lag, so when the message wasn't yet indexed the response was a header-only CSV with no data rows, which confused LLM clients into thinking the post had failed.

Drop the follow-up history call and return a short success string with channel and ts (matching ReactionsAddHandler). This removes the race entirely and saves a tier3 API call per post.

### DIFF
--- a/pkg/handler/conversations.go
+++ b/pkg/handler/conversations.go
@@ -200,7 +200,7 @@ func (ch *ConversationsHandler) UsersResource(ctx context.Context, request mcp.R
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


korotovsky:
Merging — removes a real race condition. The follow-up `conversations.history` re-fetch was racy due to Slack indexing lag; returning the channel+ts directly matches `ReactionsAddHandler` and saves an API call. Thanks!

PR #182, opened 2026-02-03
feat: add blocks parameter for raw Block Kit JSON support

sjawhar:
## Summary

Adds a new `blocks` parameter to `conversations_add_message` that allows passing raw Block Kit JSON for advanced formatting scenarios.

## Motivation

Currently, the tool converts markdown to Block Kit using `slack-go-util`. However, that library doesn't support nested lists ([PR opened to fix](https://github.com/takara2314/slack-go-util/pull/3)). This PR provides a workaround by allowing direct Block Kit JSON input.

## Changes

- Added `blocks` parameter to `conversations_add_message` tool definition
- When `blocks` is provided:
  - Takes precedence over `content_type`/`payload`
  - `payload` becomes optional fallback text for notifications/accessibility
  - JSON is parsed and validated before sending
- Updated documentation in README.md

## Use Case: Nested Lists

Markdown nested lists like:
```markdown
- Parent
  - Child 1
  - Child 2
```

Are currently stripped. With this PR, users can send Block Kit directly:
```json
{
  "blocks": "[{\"type\":\"rich_text\",\"elements\":[{\"type\":\"rich_text_list\",\"style\":\"bullet\",\"indent\":0,\"elements\":[{\"type\":\"rich_text_section\",\"elements\":[{\"type\":\"text\",\"text\":\"Parent\"}]}]},{\"type\":\"rich_text_list\",\"style\":\"bullet\",\"indent\":1,\"elements\":[{\"type\":\"rich_text_section\",\"elements\":[{\"type\":\"text\",\"text\":\"Child 1\"}]},{\"type\":\"rich_text_section\",\"elements\":[{\"type\":\"text\",\"text\":\"Child 2\"}]}]}]}]"
}
```

## Test plan

- Manual testing with Slack API
- Existing tests should pass (no breaking changes)

---
🤖 Generated with [Claude Code](https://claude.ai/code)

PR #294, opened 2026-05-07
feat: add blocks parameter and bump slack-go-util to v0.4.0

arnstarn:
## Summary

This PR adds an optional `blocks` parameter to `conversations_add_message` and bumps `takara2314/slack-go-util` from v0.3.0 to v0.4.0.

### Dependency bump: slack-go-util v0.3.0 -> v0.4.0

The v0.4.0 release (merged upstream on 2026-03-08 via https://github.com/takara2314/slack-go-util/pull/3) adds nested list support to the markdown-to-blocks converter. Without this bump, nested markdown lists (indented `- ` items) are silently dropped during conversion, producing incorrect output.

This also brings in slack-go v0.19.0 (from v0.17.3) and goldmark v1.7.16 as transitive deps.

### New `blocks` parameter

Adds an optional `blocks` string parameter to `conversations_add_message` that accepts raw Slack Block Kit JSON. When provided:

- Blocks take precedence over `text`/`content_type` for message rendering
- The `text` parameter serves as the notification fallback (as per Slack API convention)
- Either `text` or `blocks` (or both) must be provided

The parameter accepts blocks in two forms to accommodate different MCP client serialization:
- As a JSON string: `"blocks": "[{\"type\":\"rich_text\",...}]"`
- As a raw JSON array: `"blocks": [{"type":"rich_text",...}]`

### Why raw blocks alongside the markdown converter?

The text/markdown path with slack-go-util v0.4.0 handles ~90% of use cases (nested lists, bold, links, code blocks). However, there are scenarios the markdown converter cannot express:

- Blockquotes with inline formatting (converter strips bold/links inside `>` quotes)
- Mixing ordered and unordered lists at different indent levels within a single logical group
- Custom `offset` values for numbered lists (starting at a number other than 1)
- Any future Block Kit element types not yet supported by the markdown parser

Raw blocks gives users an escape hatch for these edge cases without requiring upstream changes to slack-go-util.

### Relation to PR #182

PR #182 (closed Feb 2026) proposed a similar blocks parameter. It was closed in favor of waiting for the slack-go-util nested list fix. That fix has since shipped (v0.4.0, Mar 2026) but the dependency was never bumped. This PR does both: adopts the upstream fix AND provides raw blocks for cases beyond what any markdown converter can handle.

## Test plan

- [ ] Verify `text/markdown` with nested lists produces native `rich_text_list` blocks with proper `indent` values
- [ ] Verify `blocks` parameter with raw Block Kit JSON renders correctly (bullet lists, numbered lists, nested lists, code blocks, links)
- [ ] Verify `blocks` as JSON string input works
- [ ] Verify `blocks` as raw JSON array input works
- [ ] Verify `blocks` without `text` works (blocks-only message)
- [ ] Verify `blocks` with `text` uses text as notification fallback
- [ ] Verify existing `text/plain` path is unchanged
- [ ] Verify existing `text/markdown` without nested lists still works
- [ ] Verify error handling for malformed blocks JSON

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/korotovsky/slack-mcp-server
 
-go 1.24.4
+go 1.25
 
 require (
 	github.com/gocarina/gocsv v0.0.0-20240520201108-78e41c74b4b1
@@ -13,9 +13,9 @@ require (
 	github.com/rusq/slackauth v0.7.1
 	github.com/rusq/slackdump/v3 v3.1.13
 	github.com/rusq/tagops v0.1.1
-	github.com/slack-go/slack v0.17.3
+	github.com/slack-go/slack v0.19.0
 	github.com/stretchr/testify v1.11.1
-	github.com/takara2314/slack-go-util v0.3.0
+	github.com/takara2314/slack-go-util v0.4.0
 	go.uber.org/zap v1.27.1
 	golang.ngrok.com/ngrok/v2 v2.1.1
 	golang.org/x/net v0.50.0
@@ -87,7 +87,7 @@ require (
 	github.com/ysmood/got v0.42.3 // indirect
 	github.com/ysmood/gson v0.7.3 // indirect
 	github.com/ysmood/leakless v0.9.0 // indirect
-	github.com/yuin/goldmark v1.7.13 // indirect
+	github.com/yuin/goldmark v1.7.16 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.ngrok.com/muxado/v2 v2.0.1 // indirect
 	golang.org/x/crypto v0.48.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -153,6 +153,8 @@ github.com/rusq/tagops v0.1.1 h1:R5MHPR822lSg3LFr0RS3DFS0CapRiqtuHVD5NlOMOvY=
 github.com/rusq/tagops v0.1.1/go.mod h1:mUJ5WoHxrSv9wreCrHQkAeMevt5aXFadlOdLM6UsoHc=
 github.com/slack-go/slack v0.17.3 h1:zV5qO3Q+WJAQ/XwbGfNFrRMaJ5T/naqaonyPV/1TP4g=
 github.com/slack-go/slack v0.17.3/go.mod h1:X+UqOufi3LYQHDnMG1vxf0J8asC6+WllXrVrhl8/Prk=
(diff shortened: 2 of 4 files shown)


arnstarn:
## Integration Test Failure - Not Related to This PR

The integration test failure (`TestIntegrationConversations/Test_conversations_history_tool`) is **not caused by this PR's changes**.

Evidence:
- The CI workflow uses `pull_request_target` with `actions/checkout@v6` (no ref), which checks out **master**, not the PR branch
- CI installed Go 1.24.4 (from master's `go.mod`), confirming it built master's code
- The test log shows `channels_cache_v3.json` (master's version), not `channels_cache_v2.json` (this PR's base)
- The failing test expects messages ("message 1", "message 2", "message 3") in the `#testcase-1` channel but the Slack API returns empty history
- The last successful integration run was May 2 (`design/p3-multiplexer`). No PRs have been tested between then and now, so the test workspace state likely changed independently

The `#testcase-1` channel in the test workspace may need its test messages re-populated.

arnstarn:
@korotovsky - This PR is ready for review when you get a chance. The integration test failure is unrelated (test workspace issue, see my comment above).

Happy to split this into two separate PRs if you'd prefer:
1. Dependency bump only (slack-go-util v0.3.0 -> v0.4.0) - enables nested list support in the markdown path
2. `blocks` parameter addition - allows raw Block Kit JSON for advanced formatting

Let me know if you have any questions or feedback.

korotovsky:
Merging — adds optional raw Block Kit `blocks` parameter to `conversations_add_message` (with sensible fallback to text/markdown when blocks isn't provided), and includes the `slack-go-util v0.4.0` bump that fixes nested list rendering upstream. The implementation handles both string-encoded and array-form `blocks` input robustly.

Flagging two things for visibility (not blocking the merge):
- This PR also bumps **`go 1.24.4 → go 1.25`** in `go.mod` and **`slack-go v0.17.3 → v0.19.0`** (two minor versions). Anyone building from source will need Go 1.25.
- The integration-tests CI failure appears to match the recent ngrok flake pattern; unit tests pass.

Thanks!

tomjtobinn:
I think the tool schema should advertise `blocks` as an array of objects rather than a JSON string. That lets MCP clients construct valid tool arguments directly, without hand-stringifying JSON or falling into retry loops around escaping.

It would also be worth validating that each block has a `type`. That is the minimum useful local validation: it catches obviously malformed input before calling Slack without trying to reimplement Slack’s full Block Kit validator.

Issue #115, opened 2025-10-15
Possible if we can support XOXE-XOXP token?


Sun-Phil-Kwon-Miro:
Hi team, thanks for this awesome MCP!

I think we can generate API token with refresh token in workspace

possible if we can support this option?

korotovsky:
Hi @Sun-Phil-Kwon-Miro, `xoxp` is already supported, is `xoxe` a refresh part for `xoxp`?

Sun-Phil-Kwon-Miro:
@korotovsky yeah. it seems it's a subcategory of it

the token looks like `xoxe-xoxp-<token>` it has refresher token also attached to it

korotovsky:
Normally, adding a new type of token is very time consuming task. All tools and features must be tested. Would you like to take this?

korotovsky:
Resolved by #297, which adds support for Slack's `xoxe.xoxp` refresh-token format (issued by apps with token rotation enabled). Closing as completed.

PR #297, opened 2026-05-13
fix: handle Slack refresh token format (xoxe)

aharon-br:
re https://github.com/korotovsky/slack-mcp-server/issues/115

Slack's docs https://api.slack.com/authentication/rotation and oauth.v2.access https://api.slack.com/methods/oauth.v2.access, xoxe.xoxp tokens carry have the same scopes/permissions to the regular ones. Slack apps with token rotation enabled seem to issue these instead of xoxp- tokens.


korotovsky:
Merging — closes #115. Three-line fix to recognize `xoxe.xoxp-` / `xoxe.xoxb-` token formats per Slack's token rotation docs. Thanks!

PR #308, opened 2026-05-25
fix(text): preserve URLs in rich_text link elements (match TitleLink format)

PaytonWebber:
`richTextSectionToText` drops the URL on a `RichTextSectionLinkElement` whenever the link has display text:

```go
case *slack.RichTextSectionLinkElement:
    if e.Text != "" {
        parts = append(parts, e.Text)
    } else if e.URL != "" {
        parts = append(parts, e.URL)
    }
```

That's the opposite of how `att.TitleLink` is handled, which emits `[%s](%s)` so `ProcessText`'s `normalizeLinks` can rewrite it to `url - text`.

The visible effect: machine-generated alerts (DigitalOcean, Grafana, Datadog) put their primary identifier in the link `url` and a label in `text`. The CSV column was keeping the label and dropping the identifier. For a DO alert with `url=https://cloud.digitalocean.com/droplets/562716273, text=web-prod-01` inside `"CPU alert on "`:

Before:
```
CPU alert on web-prod-01
```

After:
```
CPU alert on https://cloud.digitalocean.com/droplets/562716273 - web-prod-01
```

## Changes

`RichTextSectionLinkElement` now serializes as `[text](url)` when both fields are present and distinct. When `Text == URL` (Slack sets these identically on bare URL pastes), emit the URL once. URL-only and text-only paths are unchanged.

## Tests

Updated the existing rich-text-link case and added three: bare-URL dedup, droplet-link regression, and a new `TestBlocksToTextProcessTextPipeline` mirroring the existing `Files` pipeline test that asserts the end-to-end CSV shape.

### DIFF
--- a/pkg/text/text_processor.go
+++ b/pkg/text/text_processor.go
@@ -204,9 +204,12 @@ func richTextSectionToText(section *slack.RichTextSection) string {
 				parts = append(parts, e.Text)
 			}
 		case *slack.RichTextSectionLinkElement:
-			if e.Text != "" {
+			switch {
+			case e.Text != "" && e.URL != "" && e.Text != e.URL:
+				parts = append(parts, fmt.Sprintf("[%s](%s)", e.Text, e.URL))
+			case e.Text != "":
 				parts = append(parts, e.Text)
-			} else if e.URL != "" {
+			case e.URL != "":
 				parts = append(parts, e.URL)
 			}
 		case *slack.RichTextSectionBroadcastElement:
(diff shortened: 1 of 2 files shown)


PaytonWebber:
Closing this. After more investigation, the bug I was actually chasing (URLs stripped from DigitalOcean-style alert messages) lives on the attachment path, and it was already fixed in v1.3.0 by the `normalizeLinks` work in 6dd2feb. Upgrading resolves my case.

Issue #311, opened 2026-06-08
bin/index.js published with mode 0644 — `npx -y slack-mcp-server` fails with "Permission denied"


AndrewTKent:
### Symptom

Running `npx -y slack-mcp-server` (the install command in the README, used by most MCP clients) fails on first launch:

```
sh: /Users/<me>/.npm/_npx/<hash>/node_modules/.bin/slack-mcp-server: Permission denied
```

MCP clients wrapping this in a stdio config (Claude Desktop, Claude Code, etc.) surface it as a JSON-RPC `-32000` reconnect failure with no useful detail. The server never starts.

### Root cause

`bin/index.js` is published to the npm registry with mode `0644` instead of `0755`. The `bin` symlink points at it correctly, but the kernel refuses to exec because the `+x` bit is missing.

`npm i -g slack-mcp-server` masks the bug because the global install path applies a chmod-on-shebang heuristic to declared `bin` entries. The npx-cache path skips that, so every `npx -y` user hits this.

### Evidence

```
$ npm pack slack-mcp-server      # 1.3.0 from the registry
$ tar -tvf slack-mcp-server-1.3.0.tgz | grep bin/index
-rw-r--r--  0 0  0  2450 Oct 26 1985 package/bin/index.js
                                      ^ missing +x, epoch-zero mtime
```

For comparison, I sampled 14 other npx-cached package launchers on the same machine (`markdown-it`, `astro`, `rolldown`, `eslint`, `typescript`, etc.) — all ship `-rwxr-xr-x`. Only `slack-mcp-server` is `-rw-r--r--`.

The `Oct 26 1985` mtime in the tarball suggests something in the release pipeline is normalizing metadata with `SOURCE_DATE_EPOCH=0` (or equivalent), and that same step is clobbering the mode bit on `bin/index.js`. Common culprits: `clean-publish`, `np`, custom `tar --mtime=…` wrappers.

### Workarounds (for anyone landing here)

1. `npm i -g slack-mcp-server` and point the MCP client at the global path (e.g. `/opt/homebrew/bin/slack-mcp-server`) — global install applies `+x` correctly.
2. `chmod +x ~/.npm/_npx/<hash>/node_modules/slack-mcp-server/bin/index.js` — resets the next time npx repopulates that cache directory.

### Suggested fix

Add a `prepack` script so the bit is set right before the tarball is sealed:

```json
"scripts": {
  "prepack": "chmod +x bin/index.js"
}
```

[...]

PR #314, opened 2026-06-14
fix: make npm launcher executable

Errordog2:
## Summary

This makes the npm wrapper declared as the slack-mcp-server bin executable in git.

## Root cause

pm/slack-mcp-server/package.json maps the package binary to in/index.js, but that file is stored with mode 100644. Published tarballs preserve the non-executable mode, so 
px -y slack-mcp-server can fail with Permission denied on Unix-like systems before Node gets to run the shebang.

## Change

- Change 
pm/slack-mcp-server/bin/index.js from 100644 to 100755.
- Leave the launcher contents unchanged.

## Validation

- Verified the package in mapping points to in/index.js.
- Verified the tree entry changed from 100644 to 100755 on this branch.

Related to korotovsky/slack-mcp-server#311 and charles-openclaw/charles-microbounties#1416.


Errordog2:
CI note: `unit-tests` passed on this PR. The current `integration-tests` failure is from the workflow's OpenAI API credentials returning `429 insufficient_quota` during `TestIntegrationConversations/Test_conversations_history_tool`, not from the npm launcher mode change in this PR.

Relevant log line:

```text
POST "https://api.openai.com/v1/responses": 429 Too Many Requests ... "code": "insufficient_quota"
```

The PR changes only the git mode for `npm/slack-mcp-server/bin/index.js` from `100644` to `100755` so the package bin can execute via `npx`.

PR #316, opened 2026-06-26
feat: emit JSON instead of CSV so output pipes to jq

paymog:
## What

CLI list/table commands now print a **JSON array of objects** instead of CSV, so agents can pipe output straight to `jq`:

```sh
slack-cli channels list | jq -r '.[].Name'
slack-cli conversations history '#general' --limit 1d | jq -r '.[].Text'
slack-cli users search alice | jq -r '.[].DMChannelID'
```

Affected (tabular) commands: `channels list`/`me`, `users search`, `conversations history`/`replies`/`search`/`unreads`, `saved list`, `usergroups list`/`me`.

## How

Conversion lives entirely in the **CLI output layer** (`internal/output`), not in the shared handlers:

- `internal/cmds`: tabular commands call a new `emitTable`; everything else keeps `emit`.
- `output.Print(w, text, raw, tabular)`:
  - valid JSON (e.g. `usergroups create`) → re-indented, field order preserved;
  - `tabular` → handler's gocsv output parsed into a JSON array of string-keyed objects (column order preserved, HTML escaping off so `<@U123>`, `<https://…>` and `&amp;` stay legible);
  - plain-text status lines ("Successfully posted…", "No users found.") pass through unchanged.
- `--raw` still prints the handler's bytes verbatim — the original CSV/text — as an escape hatch.

**`pkg/handler` is deliberately untouched.** Those handlers are shared with the bundled MCP server (`cmd/slack-mcp-server`) and kept byte-for-byte with upstream for clean `git merge upstream/master`. Converting in the CLI layer keeps the MCP server's output unchanged and avoids perpetual merge conflicts, while giving the CLI the JSON it wants.

### Notes / tradeoffs

- Values are **strings** (CSV carries no type info). This avoids corrupting Slack data — e.g. message timestamps like `1772680334.954409` must not become floats. Use jq's `tonumber` for numeric comparisons. (A typed-JSON variant would require changing the shared handlers and the MCP server's contract; intentionally out of scope.)
- JSON keys are the gocsv headers, which for `Message`/`Channel` are Go field **names** (PascalCase: `Text`, `MemberCount`, `Cursor`) since those structs carry `json` tags but no `csv` tags.

## Testing

- `make test` — green; 11 new `TestUnit*` cases in `internal/output` cover CSV→JSON, column-order preservation, quoted/multiline fields, empty table → `[]`, single-column status-line pass-through, JSON re-indent, `--raw`, and the no-convert-without-tabular guard.
- `go vet` / `gofmt` clean; `cmd/slack-mcp-server` still builds.
- Verified live against a real workspace: `channels list`, `users search`, `conversations history` all emit valid JSON and pipe through `jq` (exit 0); `--raw` still returns CSV.


paymog:
Opened against the wrong base by mistake (this is a fork's CLI work); closing and reopening on paymog/slack-cli.

PR #323, opened 2026-07-08
Add conversations_info tool for per-channel metadata and read state

### LINKED ISSUE #322: Add conversations_info tool for per-channel metadata and read state
## Problem

The server can fetch a channel's messages (`conversations_history`) and mark it read (`conversations_mark`), and it can discover unreads across the workspace (`conversations_unreads`). But nothing looks up a single channel by ID to return its metadata or the caller's read state. The field that matters most is `last_read`, the timestamp of the last message the user read. That read boundary is what you need to reason about one known channel: what has arrived since you last looked.

Today `last_read` is reachable only indirectly. The `conversations_unreads` xoxp fallback (added in #171) calls `conversations.info` internally to read each channel's read state (`last_read`, or `unread_count` for DMs), but it is buried inside a capped, cross-channel scan that, on large workspaces, can silently skip a specific channel. There is no way to ask about one channel directly.

## Proposal

Add a read-only tool `conversations_info(channel_id)` that returns a single channel's or DM's metadata and the caller's read state as a one-row CSV with columns `ID`, `Name`, `LastRead`, `UnreadCount`, `IsMember`, `IsPrivate`, `IsExtShared`, `NumMembers`. It wraps the Slack Web API `conversations.info` method the server already calls internally, so it needs no new provider plumbing and no new OAuth scopes.

`last_read` is a Slack timestamp: the read boundary a client can use as a lower bound when fetching newer messages in that channel. This makes per-channel unread triage workable under OAuth user tokens (xoxp), without browser tokens or the capped cross-channel scan.

A caveat worth documenting on the tool: with xoxp tokens, `UnreadCount` is populated by Slack only for DMs; for channels, groups, and MPIMs it is 0, so callers should rely on `LastRead` rather than `UnreadCount`. With bot tokens (xoxb), Slack populates neither read-state field, so those columns come back blank/zero.

## How this differs from prior work

- #197 (closed) proposed adding `IsMember`/`HasUnreads`/`LastRead`/`MentionCount` to `channels_list`. This is narrower and complementary: a per-channel lookup by ID, not enrichment of the whole channel list.
- #229 (open) adds `activity_unreads`/`activity_mark_read` around the Activity feed. This is a direct channel lookup by ID, not the Activity feed.
- #171 (merged) added the xoxp `conversations.info` fallback inside `conversations_unreads`; this exposes that same data as a standalone, composable tool.
- It also complements #320 (an absolute `oldest` bound on `conversations_history` / `conversations_replies`): together they let a client read the boundary and fetch exactly what arrived after it. `conversations_info` is independently useful on its own as a general channel lookup.

## Example

```
conversations_info(channel_id="C0123456789")
→ ID,Name,LastRead,UnreadCount,IsMember,IsPrivate,IsExtShared,NumMembers
  C0123456789,general,1783012455.512009,0,true,false,false,42
```

gkatz2:
## Summary

Adds a read-only `conversations_info` tool that returns metadata and read state for a single channel or DM by ID: `ID`, `Name`, `LastRead`, `UnreadCount`, `IsMember`, `IsPrivate`, `IsExtShared`, `NumMembers` (a one-row CSV, matching the other read tools). It wraps the Slack `conversations.info` method, which the server already calls internally for the `conversations_unreads` xoxp fallback, so there is no new provider plumbing and no new OAuth scopes.

`LastRead` is a Slack timestamp marking the last message the caller read: the read boundary for reasoning about what has arrived in a known channel since. This gives xoxp deployments a targeted per-channel lookup instead of relying on the capped cross-channel unreads scan, which can silently skip a specific channel on large workspaces.

## Behavior

- Registered as a default-on, read-only tool and not gated by token type (like `conversations_history`). Slack still enforces the token's scopes and its access to the target conversation; the read-state fields additionally depend on token and conversation type (see the caveat below).
- `channel_id` accepts a raw channel or DM ID or a `#name`/`@user`, resolved the same way as the other conversation tools.
- Calls `conversations.info` with `IncludeNumMembers` so `NumMembers` is populated (the API omits it otherwise).
- Read-state caveat is stated in the tool description: with xoxp, `UnreadCount` is populated only for DMs; for channels, groups, and MPIMs it is 0, so rely on `LastRead`. With bot tokens (xoxb) Slack populates neither, so those columns come back blank/zero.

## Tests

- Unit test for the `slack.Channel` → output projection: full field mapping, a zero-value channel, and a DM case.
- `ValidToolNames` / tool-constant assertions extended for the new tool.

## Test plan

- Verified locally with `make build` and `make test` (unit).
- Verified read-only against a live workspace with an OAuth user token (xoxp): `conversations_info` returns `LastRead` and metadata for a channel by ID and by `#name`, and for a DM by ID and by `@handle`; a nonexistent ID returns a clear `channel_not_found` error. `NumMembers` comes back populated (confirming `IncludeNumMembers`); `UnreadCount` is 0 for channels and nonzero for a DM with unread messages (a DM with 5 unreads reported `UnreadCount` 5), matching the documented xoxp caveat.

Fixes #322


netflash:
Heads up @korotovsky — this PR is currently blocked on Trivy failing two HIGH-severity `golang.org/x/net` CVEs (CVE-2026-39821, CVE-2026-42502).

I've opened a fix targeting the branch behind this PR: https://github.com/gkatz2/slack-mcp-server/pull/1 (bumps `x/net` v0.50.0 → v0.55.0 + transitive deps, unit tests pass). It's been open ~2 weeks without a response from @gkatz2.

If it's easier, the fork with the fix applied is here: https://github.com/netflash/slack-mcp-server/tree/fix/pr323-trivy — happy to open a PR directly against this branch or against master if that's preferred.

PR #285, opened 2026-04-26
Add conversations_upload_file and conversations_download_files tools

exhenrye:
## Summary

Adds two new MCP tools to close the Slack file-operations gap:

- **`conversations_upload_file`** — upload a local file to a channel or DM, with optional `initial_comment`, `thread_ts`, and `title`. Gated by env var `SLACK_MCP_UPLOAD_FILE_TOOL`, mirroring the safety pattern of `conversations_add_message`.
- **`conversations_download_files`** — download all files attached to a given message to a local directory. Read-only, no safety gate.

## Motivation

File upload/download is the most common MCP gap I've hit running this server day-to-day — sending PDFs / fetching attachments otherwise requires shell helpers or out-of-band tooling. Both tools follow existing conventions in the codebase (handler structure, env-gating, provider abstraction).

## Provider changes

- `UploadFileV2Context` added to the `SlackAPI` interface (used by the upload tool).
- `Token()` and `Cookies()` accessors exposed on `MCPSlackClient`; `ApiProvider` gains `HTTPAuth()` for raw HTTP file downloads, handling both bot-token (`xoxb` / `xoxp`) and stealth (`xoxc` / `xoxd`) auth uniformly.

## Testing

Smoke-tested end-to-end against a live workspace with both bot-token and stealth (`xoxc` / `xoxd`) auth — upload + download round-trip works for images, PDFs, and text files; thread replies and `initial_comment` honored.

## Note on base

This branch is based on `5637b6f` rather than current `master` — happy to rebase onto current `master` if there's interest in merging. Wanted to surface the contribution first before doing the conflict-resolution work, since `pkg/handler/conversations.go`, `pkg/provider/api.go`, and `pkg/server/server.go` have all seen substantial change since the fork point.

Issue #247, opened 2026-03-23
Support file upload and snippet sharing


chriscoey:
Currently the server supports downloading file attachments via `attachment_get_data`, but there's no way to upload files or share Slack file/snippet objects.

**Use cases:**
- Share generated code snippets or script output as Slack file objects (syntax highlighting, collapsible)
- Upload CSV/JSON data files as part of a conversation
- Post log files or error traces for debugging discussions

**Prior art:**
PR #121 added file upload support via `UploadFileV2Context` (the current Slack v2 upload flow). It was closed after requested security controls — default-off gating and allowlisting of accessible filesystem paths and remote domains — were not added.

**Possible approach (open to other ideas):**
- New tool, e.g. `file_upload`, gated behind `SLACK_MCP_FILE_UPLOAD_TOOL` env var (disabled by default), consistent with existing write tool gating
- Source allowlists for safety: allowed local path patterns and/or allowed remote domains, addressing the security concerns from #121. Optional extras like max size or target-channel restrictions
- Accept content as text (with optional `snippet_type`) or base64 for binary, mirroring how `attachment_get_data` returns content

**Auth note:** Slack's upload APIs require `files:write` scope, so this would likely only work with OAuth/bot tokens, not browser-token mode. Worth documenting up front.

This would complete the round-trip with the existing `attachment_get_data` download support.

PR #334, opened 2026-07-29
feat: add file_upload tool (works with browser session tokens)

MaksBoi:
## Summary

Adds the `file_upload` tool proposed in #247. The server could already download attachments (`attachment_get_data`) but had no way to send anything back.

**It also works in stealth mode.** This contradicts the assumption that killed earlier attempts, and I have receipts, see below.

Content comes from exactly one of three sources:

- `file_path` for a file on disk
- `content` for UTF-8 text (logs, snippets)
- `content_base64` for binary data, so an agent can post a screenshot it holds in memory without touching the filesystem

Files land in a channel or a thread through `files.getUploadURLExternal` → upload → `files.completeUploadExternal` (`UploadFileContext` in slack-go v0.19.0), with `initial_comment` as the caption.

## Browser tokens do work

The proposal in #247 says upload "would likely only work with OAuth/bot tokens, not browser-token mode", and #309 rejects `xoxc`/`xoxd` at call time because `files.*` "does not document support for them". Both are assumptions, not measurements. I ran the full three-step flow against a live workspace on `xoxc` + `d` cookie:

```
auth.test               ok
getUploadURLExternal    ok, file_id + upload_url returned
POST body               HTTP 200
completeUploadExternal  ok, file shared with a permalink
files.delete            ok
```

Then the same round trip through this branch: a 2.6MB PNG from disk, a text file from `content`, and a PNG from `content_base64`. All three arrived, both images rendered with previews.

Nothing extra is needed for this. The `xoxc` client is already built against the team endpoint (`api.go`), and `transport.go` attaches the `d` cookie to every request including `files.slack.com`, so the existing plumbing carries it.

One quirk worth documenting: Slack strips PNG metadata chunks (`pHYs`, `tEXt`) on upload, so the stored file is a few dozen bytes smaller than the original. Pixel data (`IDAT`) is byte-identical, nothing is corrupted.

## Security

#121 and #285 were closed for missing controls. All of them are here.

**Tool is off by default.** Not registered unless `SLACK_MCP_FILE_UPLOAD_TOOL` is set. That variable doubles as a channel allowlist with the same semantics as `SLACK_MCP_ADD_MESSAGE_TOOL`: `true`, a CSV of channel IDs, or `!C123` to exclude.

**Path allowlist.** `file_path` is rejected outright unless `SLACK_MCP_FILE_UPLOAD_PATHS` lists the directories the server may read. Paths go through `EvalSymlinks` first, then `filepath.Rel` decides containment, so `/data/allowed` does not authorise `/data/allowed-evil`.

**No TOCTOU window.** Validating a path and reopening it by name later is exploitable: the name can point somewhere else by the time you read it. I measured this on a draft of the patch, and under an active rename race 63% of the calls that passed validation read the swapped file. So the file is opened during validation with `O_NOFOLLOW|O_NONBLOCK`, re-checked through the descriptor, and every later step uses that descriptor. `O_NONBLOCK` plus the `IsRegular` check also stops a FIFO from hanging the handler forever, which plain `os.Open` does with no way to cancel it.

**Channel resolved before the policy check**, so `#general` cannot reach a channel the allowlist excludes.

**Request params are not logged.** For this tool they carry the file payload itself.

**No domain allowlist, because there is no URL fetching.** Remote sources are deliberately out of scope. If fetching by URL is ever wanted, that allowlist should land with it.

## Notes

- Uploads stream from the descriptor, so the 100MB cap does not mean 100MB of memory.
- Open flags live in build-tagged files, Windows still builds. Verified across `darwin`, `linux`, `windows` × `amd64`, `arm64`.
- Tests sit next to the handler (`file_upload_test.go`, `file_upload_open_unix_test.go`), 7 test functions, ~35 subtests: allowlist escapes, symlink and traversal cases, the sibling-prefix trap, FIFO and device rejection, source validation, size caps, filename sanitising, channel policy.
- `TestValidToolNames` in `server_test.go` needed the new constant. Worth knowing: `make test` filters on `-run=".*Unit.*"`, so that test never runs there and the breakage stays invisible until someone runs `go test ./...`.
- Rebased on current `master` (`b88c0de`).

## Overlap with #309

#309 is open and covers similar ground. The differences that matter: this branch supports browser tokens instead of rejecting them, closes the TOCTOU window at open time rather than reading the whole file into memory, and rejects FIFOs and devices. Happy to fold this into #309 instead if you prefer that route.


anaisbetts:
Would be very useful!

Issue #145, opened 2025-12-24
Markdown code blocks render as inline code instead of code blocks


moltar:
## Problem

When posting messages with `content_type: "text/markdown"`, multiline fenced code blocks are incorrectly rendered as separate inline code snippets instead of a single code block.

### Example

Input markdown:
~~~
```
foo
bar
```
~~~

**Expected**: A single multiline code block in Slack

**Actual**: Two separate inline code snippets: `foo` `bar`

## Root Cause

The upstream library `github.com/takara2314/slack-go-util` (v0.2.0) incorrectly converts fenced code blocks.

- **Current behavior**: Uses `RichTextSection` with `style: {code: true}` which creates inline code formatting
- **Correct behavior**: Should use `RichTextPreformatted` which creates proper preformatted code blocks

## Location in this codebase

`pkg/handler/conversations.go:174`:
```go
blocks, err := slackGoUtil.ConvertMarkdownTextToBlocks(params.text)
```

## Fix

PR submitted to upstream: https://github.com/takara2314/slack-go-util/pull/1

## Workaround

Until the upstream fix is merged and a new version released, users can:
1. Use `content_type: "text/plain"` and format code blocks manually using Slack's mrkdwn syntax
2. Or pin to a fork with the fix applied

moltar:
That upstream PR has been merged. We need to update the version used here.

korotovsky:
Ref: https://github.com/korotovsky/slack-mcp-server/pull/147

Issue #140, opened 2025-12-18
conversations_add_message does not render Slack mrkdwn formatting (bold, mentions, etc.)


joeyfrasier:
# GitHub Issue for korotovsky/slack-mcp-server

**Title:** `conversations_add_message` does not render Slack mrkdwn formatting (bold, mentions, etc.)

## Issue Summary
The `conversations_add_message` tool does not properly render Slack's native `mrkdwn` formatting when posting messages. Both text/plain and text/markdown content types result in literal text output instead of formatted Slack messages.

## Expected Behavior
When posting messages with Slack formatting syntax, the formatting should render properly:
- Bold text: `*bold*` should render as **bold**
- User mentions: `<@USER_ID>` should render as clickable user mentions
- User group mentions: `<!subteam^GROUP_ID>` should render as clickable group mentions

## Actual Behavior
Formatting is either stripped or displayed as literal text:
- Bold text appears as plain text
- User group mentions display as `<@subteamGROUP_ID>` (visible angle brackets and literal text)
- User mentions may display similarly

## Steps to Reproduce
```json
{
  "channel_id": "#test-channel",
  "content_type": "text/plain",
  "payload": "*Project Name:* This is a test message cc: <@U1234567> <!subteam^S040YBMF9B5>"
}
```

Result: Message posts with literal text including asterisks and angle brackets, no formatting applied.

## Environment
- MCP Server: korotovsky/slack-mcp-server (latest)
- Token type: xoxp (user token)
- Transport: stdio

## Additional Context
According to Slack's API documentation, messages should use `mrkdwn` format with proper syntax:
- User group mentions: `<!subteam^ID|@groupname>` ([Slack docs](https://api.slack.com/reference/surfaces/formatting#mentioning-groups))
- The Slack API's `chat.postMessage` method supports `mrkdwn` via the `mrkdwn` parameter

## Suggested Solution
The `conversations_add_message` tool should either:
1. Pass formatting through properly to Slack's API with `mrkdwn: true`
2. Document the limitation and provide guidance on supported formatting
3. Add a `mrkdwn` parameter to enable/disable formatting support

---

## How to Post This Issue

1. Go to: https://github.com/korotovsky/slack-mcp-server/issues/new
2. Copy the title and description above
3. Submit the issue

korotovsky:
It is likely issue in github.com/takara2314/slack-go-util.

PR #339, opened 2026-08-15
fix(add-message): document the Markdown dialect and constrain content_type

dazebug:
## Summary

`conversations_add_message` converts `text` with `slack-go-util` whenever `content_type=text/markdown` (the default). That library runs goldmark **without** GFM extensions, so the dialect it actually accepts is bare CommonMark — but nothing in the tool schema said so. In a Slack context that silence is actively misleading: Slack's own format is called *mrkdwn*, so a caller's natural prior is "posting to Slack means Slack syntax."

Writing mrkdwn does not fail loudly. It corrupts or deletes content and returns success.

This PR writes the contract into the schema and the README. **No handler or parser behavior changes.**

## Measured failure modes

Running the pinned `takara2314/slack-go-util v0.4.0` directly on each input:

```
"see <https://example.com|label> now"  ->  "see  now"          link DELETED
"see <https://example.com> now"        ->  "see  now"          link DELETED
"• alpha\n• beta"                      ->  "• alpha• beta"     one paragraph, joined with no separator
"line one\nline two"                   ->  "line oneline two"  same
"*bold intent*"                        ->  "_bold intent_"     level-1 emphasis becomes Slack italic
"| a | b |\n| - | - |"                 ->  "| a | b || - | - |" pipes are literal, no GFM tables
"- [ ] todo"                           ->  bullet "[ ] todo"   no GFM task lists
"~~gone~~"                             ->  "~~gone~~"          no GFM strikethrough
"- cc <@U123>"                         ->  rich_text text node, mention does not resolve
```

Why the deletions happen: `<url|label>` and `<url>` are valid CommonMark autolinks, and `convertInlineMarkdownToMrkdwn` handles only `Text`, `Emphasis`, `Link` and `CodeSpan` nodes. Every other kind recurses into its children — and an autolink has none, so its content is dropped.

Nothing catches this. Every branch of the converter's `ast.Walk` callback returns a nil error, so the handler's "markdown parse failed, fall back to plain text" path is unreachable for these inputs.

Mentions are container-dependent, which is worth stating explicitly: a paragraph becomes a `section` block with an `mrkdwn` text object and resolves `<@U123>`, while a list item or block quote becomes `rich_text` and a heading becomes `plain_text`, neither of which does.

## Changes

- The tool description and the `text`, `content_type` and `blocks` descriptions now name the dialect, the destructive cases, the reliable forms, and the `text/plain` escape hatch.
- `content_type` gains `mcp.Enum("text/markdown", "text/plain")` so schema-aware clients constrain the value up front. The parser's runtime validation stays, since mcp-go does not validate arguments before dispatch.
- The garbled inherited `thread_ts` sentence ("…or a message in the thread_ts must be the timestamp…") is repaired in both the schema and the README; it sits in the same hunk.
- README listed `payload` as required and omitted `text` and `blocks` entirely, documenting a schema the server no longer exposes. It now matches what the tool advertises. `payload` stays an undocumented backward-compatibility alias so new callers do not adopt it.
- The definition moves into `newConversationsAddMessageTool()` purely for testability: `NewMCPServer` authenticates against Slack before it returns, so the registered schema could not be asserted from a unit test.
- New `pkg/server/tools_test.go` pins the contract — the enum, the default, property types, the absence of `payload`, and substring assertions for each rule so wording stays editable while the rules do not. The tests carry the `TestUnit` prefix so `make test` actually runs them; without it the `-run=".*Unit.*"` filter would skip the whole file.

## Out of scope

Server-side detection or normalization of mrkdwn input, block splitting for Slack's 3000-character section limit and 50-block cap, and any change to the converter itself. Those are behavior changes and belong in their own PR.

## Why document the converter instead of replacing it

#330 proposes rendering `text/markdown` through Slack's native `markdown` block instead of the converter, which would make tables and headings render properly. It is the obvious objection to this PR: why write down quirks that someone is trying to delete? Three reasons this still belongs on its own.

**They fix different halves of the bug.** That PR changes what the server *does*; this one changes what the server *says*. Adopting the native block does not resolve the ambiguity that caused the incident — the word "markdown" next to the word "Slack". It relocates it: by that PR's own description, the native block interprets GFM, so `*bold*` renders as *italic* there too, the opposite of the mrkdwn a caller was reaching for. Someone who writes Slack syntax is still wrong afterwards, just wrong in a different way, and the schema still has to say which dialect it means.

**It makes an explicit contract more necessary, not less.** That PR keeps `ConvertMarkdownTextToBlocks` as a fallback for input over the `markdown` block's 12,000-character limit. That leaves two rendering paths with different dialects behind one parameter — precisely the situation a schema description exists to disambiguate.

**The overlap is cheap to unwind.** If that approach or an equivalent lands, the failure inventory here (autolink deletion, paragraph joining, missing GFM constructs) has to be rewritten. That edit is confined to one function, and the regression test added here is what makes it safe to perform. Everything else survives untouched: the `content_type` enum, the README correction, the `thread_ts` repair, and the principle that the tool must name its dialect.

Since this PR changes no behavior, the two cannot conflict semantically — only textually.

## Related work

- #140 (open issue) is this same confusion reported from the user side; the maintainer's reply points at slack-go-util.
- #145 (closed issue, code blocks rendered as inline code) and #147 (merged, a dependency bump aimed at multiline handling) are earlier rounds of the same converter-fidelity problem. The paragraph joining documented above is still present in v0.4.0.
- #294 (merged) added `blocks` and pinned slack-go-util v0.4.0 — it is the provenance of the `blocks` and `content_type` strings this PR replaces.
- #330 (open) proposes replacing the converter with Slack's native `markdown` block; see the section above for how the two relate.
- #334 (open) adds `file_upload`, whose `initial_comment` passes straight to Slack with no converter and is correctly described as supporting Slack mrkdwn. The two tools genuinely speak different dialects — one more reason for each to say which one it means.

**Rebase note:** #312 (open) inserts a `reply_broadcast` property inside this same tool block. Because this PR moves the block into `newConversationsAddMessageTool()`, the two will conflict textually; resolving it means adding the new property inside the extracted function.

## Testing

`make test` (the unit-test job's command) passes, as do `go test ./pkg/server/`, `go vet ./pkg/server/` and `gofmt`. The `pkg/handler` integration tests are unaffected and still require `SLACK_MCP_XOXP_TOKEN` / `SLACK_MCP_OPENAI_API`.


dazebug:
A note on the red check, so it is not read as a regression from this PR.

`integration-tests` fails on `TestIntegrationConversations/Test_conversations_history_tool` with `There are no messages found in the channel #testcase-1. The channel appears to be empty.` — the fixture channel in the test workspace has no messages. The same job failed on the previous eight PR runs across unrelated branches (`fix/resilient-users-cache-fetch`, `pr-scheduled-messages`, `pr-replycount`, `feat/file-upload`, `feat/canvas-tools`, `feat/markdown-native-block`), so it predates this branch.

Worth knowing either way: both workflows use `pull_request_target` with `actions/checkout` and no `ref`, so they check out the base branch rather than the PR head. In run `31907489580` the checkout step fetched `b88c0de` (master), and `pkg/server` reported `[no tests to run]` — master's state, since these tests only exist on this branch. Neither the green nor the red check reflects this diff.

Locally on this branch: `make test` passes with both new `pkg/server` tests executing, as do `go vet ./pkg/server/` and `gofmt`.

PR #353, opened 2026-09-09
fix: honour unlisted 1/min history and replies quotas

paymog:
`conversations history` and `conversations replies` now survive the unlisted-app special tier (1 req/min, 15 messages/page) instead of 429ing inside the old 30s timeout.

A long `Retry-After` is remembered on disk so later processes cap the page and share the slot. Force that mode with `SLACK_MCP_UNLISTED_HISTORY=1` (Sinatra should). Default `--timeout` is 2m so one wait fits. Paginate with `--cursor` for the rest of the thread.

`go test ./pkg/limiter/` covers cap, wait, 429 retry, and the process lock.


paymog:
Opened against the wrong repo. This change belongs on paymog/slack-cli.

